PR #8, opened 2021-04-13
Add dummy POCOs for Bash, PowerShell and Publish steps

premun:


mandel-macaque:
I want to wait to do the review based on what we decide regarding #9

premun:
@mandel-macaque I have moved the optional parameters out of ctors. Required are usually just names and display names.

We could check this in if you agree and I would start building the API.

src/Sharpliner.Model/Job.cs
```diff
@@ -3,20 +3,40 @@
...
-        IEnumerable<Step> Steps,
-        JobWorkspace Workspace = JobWorkspace.Outputs,
-        bool ContinueOnError = false);
 
     // TODO: missing properties Uses, Services
+    public record Job
+    {
+        public string Name { get; }
+
+        public string DisplayName { get; }
```
mandel-macaque, line 12: Nullable? That will help us to validate.

```suggestion
        public string? DisplayName { get; }
```

src/Sharpliner.Model/Job.cs
```diff
@@ -3,20 +3,40 @@
...
+
+        public ContainerReference? Container { get; init; }
+
+        public TimeSpan? Timeout { get; init; }
+
+        public TimeSpan? CancelTimeout { get; init; }
+
+        public IEnumerable<Variable>? Variables { get; init; }
+
+        public IEnumerable<Step>? Steps { get; init; }
```
mandel-macaque, line 30: This is going to have to be a list and be initialize already. That way users just have to use {} which results in a call to .Add
```suggestion
        public List<Step> Steps { get; } = new List<Step>();
```
If you try the API as you wrote it the user will have to do:
```csharp 
Steps = new [] {}
// or
Steps = new List<Step> ();
// But if the list is already present
Steps = {
new Step()
}
```

src/Sharpliner.Model/Job.cs
```diff
@@ -3,20 +3,40 @@
...
+
+        public JobWorkspace Workspace { get; init; } = JobWorkspace.Outputs;
+
+        public bool ContinueOnError { get; init; } = false;
+
+        public Job(string name, string displayName)
+        {
+            Name = name ?? throw new ArgumentNullException(nameof(name));
```
mandel-macaque, line 38: So we need that with nullable? I mean, with and older version of C# the API is horrible...

src/Sharpliner.Model/Pipeline.cs
```diff
@@ -2,10 +2,16 @@
...
-        IEnumerable<Variable> Variables,
-        IEnumerable<Trigger> Triggers,
-        IEnumerable<Stage> Stages);
+    public record Pipeline
+    {
+        public string? Name { get; init; }
+
+        public Resources? Resources { get; init; }
+
+        public IEnumerable<Variable>? Variables { get; init; }
```
mandel-macaque, line 11: Same, List and init it

src/Sharpliner.Model/Pipeline.cs
```diff
@@ -2,10 +2,16 @@
...
+
+        public IEnumerable<Trigger>? Triggers { get; init; }
```
mandel-macaque, line 13: Same

src/Sharpliner.Model/Pipeline.cs
```diff
@@ -2,10 +2,16 @@
...
+
+        public IEnumerable<Stage>? Stages { get; init; }
```
mandel-macaque, line 15: Same

src/Sharpliner.Model/Pool.cs
```diff
@@ -0,0 +1,23 @@
...
+        internal PoolName(string name) : base()
+        {
+            Name = name ?? throw new ArgumentNullException(nameof(name));
+        }
+    }
+
+    public record HostedPool(
+        string Name,
+        string VmImage,
+        IEnumerable<string>? Demands = null)
```
mandel-macaque, line 21: Move demands to a property too

src/Sharpliner.Model/Stage.cs
```diff
@@ -2,11 +2,24 @@
...
-        string Condition,
-        IEnumerable<Variable> Variables,
-        IEnumerable<Job> Jobs);
+    public record Stage
+    {
+        public string Name { get; }
+
+        public string DisplayName { get; }
+
+        public IEnumerable<string>? DependsOn { get; init; }
```
mandel-macaque, line 11: Same comment about the list

src/Sharpliner.Model/Stage.cs
```diff
@@ -2,11 +2,24 @@
...
+
+        public string? Condition { get; init; }
+
+        public IEnumerable<Variable>? Variables { get; init; }
```
mandel-macaque, line 15: Same

mandel-macaque (changes requested):
Small changes about the collections.   made the same mistake on my PR and I noticed because I wrote the test and saw the extra tying. I know it looks stupid, but can you write some dummy tests (ex: throws NRE). They will
Increase coverage, but best of all, we can see the API in use.

src/Sharpliner.Model/Job.cs
premun: Oh my god! This is amazing! I didn't know about this! I was always wondering how we were gonna achieve `{}` only

src/Sharpliner.Model/Job.cs
premun: If someone who doesn't use nullable uses our library, he can pass null in. Or someone else could make a mistake and pass null even when they have nullability on.

I am not 100% sure display name is actually needed by AzDO but maybe we should require it?

premun:
@mandel-macaque I am moving these to the AzDO namespace in the API branch

src/Sharpliner.Model/Job.cs
premun: I don't understand?

mandel-macaque (approved):

PR #13, opened 2021-04-17
[Mode][GitHub] Add initial implementation of the triggers.

mandel-macaque:
Split the model namespace in two:

- AzDO
- GitHub

And provide the initial work for the GitHub actions model. At the moment
we have just the Workflow (is a pipeline in the AzDO world) triggers.

GitHub actions provides triggers for all its webhooks and a few manual
ones as well as a cron job. Tests have been added monstly to show the
API and make sure we are not doing anything stupid.

Reference: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#configuring-workflow-events


src/Sharpliner.Model/GHActions/Webhook.cs
```diff
@@ -0,0 +1,450 @@
...
+    // }
+    //
+    // we will have to wait until c# does implement the feature, so for the time
+    // being we will do:
+    //
+    // var cr = new CheckRun {
+    //     Activities = { CheckRun.Activity.Created, CheckRun.Activity.Requested }
+    // }
+    //
+    // verbose, but type safe and autocomplete will help. If not, get a better editor!
```
mandel-macaque, line 26: @premun this will make  it very nice, they are working on it.

premun:
I have named the namespace `AzureDevOps` in my git branch

tests/Sharpliner.Model.Tests/GitHub/WorkflowTests.cs
```diff
@@ -0,0 +1,154 @@
...
+            };
+            Assert.NotNull(w);
+            Assert.Null(w.On.Manuals.WorkflowDispatch);
+            Assert.Null(w.On.Manuals.RepositoryDispatch);
+            Assert.Empty(w.On.Schedules);
+            Assert.Single(w.On.Webhooks);
+        }
+
+        [Fact]
+        public void Workflow_Several_webhook()
```
premun, line 127: I am not sure unit tests like this one won't just add extra work to any refactoring which we might do in the early stages?

src/Sharpliner.Model/GHActions/Webhook.cs
```diff
@@ -0,0 +1,450 @@
...
+        public enum Activity
+        {
+            Completed,
+            Requested,
+        }
+
+        /// <summary>
+        /// List with the workflows that will trigger the event.
+        /// </summary>
+        [Required]
```
premun, line 436: Where does this attribute come from?

src/Sharpliner.Model/GHActions/Webhook.cs
premun: Nice, also didn't know this is coming!

premun: We could potentially live on top of net6.0 and have the `LangVersion` set to `latest` and get this early :)

premun (approved):
Nice! You are fast!

tests/Sharpliner.Model.Tests/GitHub/WorkflowTests.cs
mandel-macaque: I like them because they allow me to test the API... and I want to also test the static analysis using the tests :) I don't mind doing it.

src/Sharpliner.Model/GHActions/Webhook.cs
mandel-macaque: From the BCL but we might want to add our own in the near future. Using it atm but maybe removing it might be a good idea.

mandel-macaque:
@premun merged with main, so it should be everything ready to land :)

PR #14, opened 2021-04-18
[Model] Add conditional variable definition

premun:
This adds the possibility of adding variables and groups and making nested if statements where groups of variables will share the condition:
```csharp
public override Pipeline Pipeline => new()
{
    Variables =
    {
        Variable("Configuration", "Release"),
        Group("PR keyvault variables"),

        If.Equal(variables["Build.Reason"], "PullRequest")
            .Variable("TargetBranch", "$(System.PullRequest.SourceBranch)")
            .Variable("IsPr", true),

        If.And(Equal(variables["Build.SourceBranch"], "refs/heads/production"), NotEqual("Configuration", "Debug"))
            .Variable("PublishProfileFile", "Prod")
            .If.NotEqual(variables["Build.Reason"], "PullRequest")
                .Variable("AzureSubscription", "Int")
                .Group("azure-int")
            .EndIf
            .If.Equal(variables["Build.Reason"], "PullRequest")
                .Variable("AzureSubscription", "Prod")
                .Group("azure-prod"),
    }
};
```

I don't have the serializer for it yet but since the definitions have parent pointers, it should be easy to reach the root and build the YAML from there.

We can also then transform it into the full declaration language easily (have `each`...) but I don't believe we really need `each` because it can be handled in C# and generate the definitions programmatically there


src/Sharpliner.Model/ConditionedDefinition.cs
```diff
@@ -0,0 +1,139 @@
+﻿using System;
```
premun, line 1: I should add comments to this generics hell

premun:
We have one last problem - we can currently do `If` in the `Variable` section but it will clash when trying to use it in the other sections. We only have one `If` for everything.

tests/Sharpliner.Serialization.Tests/MockPipeline.cs
```diff
@@ -0,0 +1,37 @@
...
+        public override TargetPathType TargetPathType => TargetPathType.RelativeToGitRoot;
+
+        public override Pipeline Pipeline => new()
+        {
+            Name = "$(Date:yyyMMdd).$(Rev:rr)",
+
+            Variables =
+            {
+                new Variable("Configuration", "Release"), // We can create the objects and then resue them for definition too
+                Variable("Configuration", "Release"),     // Or we have this more YAML-like definition
```
mandel-macaque, line 18: I dig this version, but I'm sure you knew.

tests/Sharpliner.Serialization.Tests/MockPipeline.cs
```diff
@@ -0,0 +1,37 @@
...
+                Group("PR keyvault variables"),
+
+                If.Equal("variables['Build.Reason']", "PullRequest")
+                    .Variable("TargetBranch", "$(System.PullRequest.SourceBranch)")
+                    .Variable("IsPr", true),
```
mandel-macaque, line 23: Static analysis fo the variable names to be unique is going to be interesting, but is doable with this syntax.

tests/Sharpliner.Serialization.Tests/MockPipeline.cs
```diff
@@ -0,0 +1,37 @@
...
+
+                If.And(Equal("variables['Build.SourceBranch']", "refs/heads/production"), NotEqual("Configuration", "Debug"))
```
mandel-macaque, line 25: Can we have:
```suggestion
                If.And(Equal(Variable(Build.SourceBranch"), "refs/heads/production"), NotEqual("Configuration", "Debug"))
```
We should be able to infer that is a variable and later do write the 'variables['Build.SourceBranch'] inside the if model.

mandel-macaque (changes requested):
Small comment to make it even nicer :)

tests/Sharpliner.Serialization.Tests/MockPipeline.cs
premun: I want to keep both so that you can define variables in C# and then just drop them in Jobs/Steps so you can use the benefit of being in the C# program. The same needs to happen for other pieces of the pipeline.

I can imagine that you define your own steps like "Signing Job" as C# objects and then assemble pipelines out of them. The pipeline then could be just 10 lines of objects and super clear to understand

tests/Sharpliner.Serialization.Tests/MockPipeline.cs
premun: We can even have 
```
Equal(variables["Build.SourceBranch"], "refs/heads/production")
```

What do you say?

tests/Sharpliner.Serialization.Tests/MockPipeline.cs
premun: I went through some of our YAML files and we always group more definitions under one conditions so I wanted to have that here

tests/Sharpliner.Serialization.Tests/MockPipeline.cs
premun: @mandel-macaque I updated the `variables[]` usage - now it's almost 100% like YAML :)

premun:
Weird, why doesn't this PR show https://github.com/premun/sharpliner/commit/32db9c892084078bf98657a522e117e1811b8343 ???

mandel-macaque (approved):

mandel-macaque:
Merge o/ o/ o/!!

PR #17, opened 2021-05-04
Prevent mixing different types of definitions

premun:
You won't get Intellisense for `.Stage()` in a variable block


premun:
Covered in #18

PR #19, opened 2021-05-05
[Model] GithubAction - Add permissions to the workflow.

mandel-macaque:
Allow to set the persmissions for a workflow, there are several ways to
use the API.

1. Provide each scope we want to Read or Write

```csharp
Permissions =
{
    Write =
    {
        GitHubPermissionScope.Actions,
    },
    Read = {
        GitHubPermissionScope.Actions,
        GitHubPermissionScope.Contents
    }
}
```

2. Read All

Use the fluent API

```csharp
Permissions = new Permissions().All(GitHubPermission.Read)
```

3. Write All

Use the fluent API

```csharp
Permissions = new Permissions().All(GitHubPermission.Write)
```

4. Read and Write All

Use the fluent API

```csharp
Permissions = new Permissions().All(GitHubPermission.Write).All(GitHubPermission.Read)
```

5. A mix of Read/Write All and more specific

Use a mix of declarative + fluet:
```csharp
Permissions = new Permissions
{
  Read =
  {
    GitHubPermissionScope.Actions,
  }
}.All(GitHubPermission.Write)
```

The above API will also be used for jobs.

### DIFF
--- a/src/Sharpliner.Model/GHActions/Permissions.cs
+++ b/src/Sharpliner.Model/GHActions/Permissions.cs
@@ -1,0 +1,56 @@
+using System;
+using System.Collections.Generic;
+
+namespace Sharpliner.Model.GHActions
+{
+    public enum GitHubPermissionScope
+    {
+        Actions,
+        Checks,
+        Contents,
+        Deployments,
+        Issues,
+        Packages,
+        PullRequests,
+        RepositoryProjects,
+        SecurityEvents,
+        Statuses,
+    }
+
+    public enum GitHubPermission
+    {
+        None,
+        Read,
+        Write
+    }
+
+    public record Permissions
+    {
+        public HashSet<GitHubPermissionScope> Read { get; } = new();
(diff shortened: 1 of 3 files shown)


src/Sharpliner.Model/GHActions/Permissions.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        None,
+        Read,
+        Write
+    }
+
+    public record Permissions
+    {
+        public HashSet<GitHubPermissionScope> Read { get; } = new();
+        public HashSet<GitHubPermissionScope> Write { get; } = new();
```
mandel-macaque, line 30: I'm going to have some fun serialising this, but the API is so much nicer.

premun: Serializing is a piece of cake actually. Look up `IYamlConvertible` in our code and see how it is implemented.

premun (approved):

PR #26, opened 2021-05-12
[Model] Add env to github workflow.

mandel-macaque:


src/Sharpliner/GitHubActions/Workflow.cs
```diff
@@ -39,5 +40,11 @@ public record Workflow
         /// setting will apply to all the jobs in a workflow. You can override this setting per job.
         /// </summary>
         public Permissions Permissions { get; init; } = new ();
+
+        /// <summary>
+        /// A map of environment variables that ara available to the steps of all jobs. When more than one variable
```
premun, line 45:
```suggestion
        /// A map of environment variables that are available to the steps of all jobs. When more than one variable
```

src/Sharpliner/GitHubActions/Workflow.cs
```diff
@@ -39,5 +40,11 @@ public record Workflow
...
+        /// with the same name is used, the most specific one will be used.
```
premun, line 46:
```suggestion
        /// with the same name is used, the latter one will be used.
```

tests/Sharpliner.Tests/GitHub/WorkflowTests.cs
```diff
@@ -291,5 +291,48 @@ public void Workflow_Detailed_Permissions()
...
+                    {
+                        new PullRequest
+                        {
+                            Activities = { PullRequest.Activity.Assigned, PullRequest.Activity.Closed }
+                        }
+                    }
+                },
+                Enviroment =
+                {
+                    ["Database"] = "production",
```
premun, line 331: Ah cool, I will change my env definitions to this style.

We should then also give some styleguide so that people also write clean definitions

premun (approved):

tests/Sharpliner.Tests/GitHub/WorkflowTests.cs
mandel-macaque: I saw yours, I liked it. But then I make lazy lol

PR #27, opened 2021-05-12
[Model] Provide concurrency for github workflows.

mandel-macaque:


src/Sharpliner/GitHubActions/Concurrency.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+
+namespace Sharpliner.GitHubActions
+{
+    public record Concurrency
+    {
+        /// <summary>
+        /// Represents a concurrency group that will be used to coordinate the execution of the workflow. This will
+        /// ensure that only one workflow is executed at a specific time. If
```
premun, line 9:
```suggestion
        /// ensure that only one workflow is executed at a specific time.
```

premun (approved):

PR #29, opened 2021-05-12
Adds publishing check and makes Sharpliner self-hosted

### LINKED ISSUE #3: Guardian build task
Every time we change the C# definition of the pipeline, it needs to be exported into the `.yaml` file in the repo
which the pipeline is defined by. I currently don't see a way to have the C# definition only.

There should be a build step defined that will run in PRs and verify that the C# definition was exported into the YAML file in the repo.
It will export the definition and compare with the YAML file and fail if it doesn't match.
This way you will have a validation that you did the export before opening a PR.

premun:
**Makes the project use the pipeline defined using Sharpliner inside the `Sharpliner.CI` project!**

Adds an AzDO task that will verify whether you have published and commited all changes into YAML files. (resolves #3)

The many empty arrays in the YAML will be gone once we check in https://github.com/aaubry/YamlDotNet/pull/608


premun:
@mandel-macaque I changed YAML and named one stage "Build and test 2", the number 2 is not in the C# definition of the pipeline -> the build fails :)

So we will have the build check whether we committed all of the changes always

mandel-macaque (approved):

PR #31, opened 2021-05-17
[Model] Add jobs to github action model.

mandel-macaque:
Few things need documentation. I need to see how they'll work. Specially with the If condition.

### DIFF
--- a/src/Sharpliner/GitHubActions/Environment.cs
+++ b/src/Sharpliner/GitHubActions/Environment.cs
@@ -1,0 +1,20 @@
+using System;
+
+namespace Sharpliner.GitHubActions
+{
+    /// <summary>
+    ///
+    /// </summary>
+    public record Environment
+    {
+        public Environment(string name, string? url = null)
+        {
+            if (string.IsNullOrEmpty(name))
+                throw new ArgumentNullException(nameof (name));
+            Name = name;
+            Url = url;
+        }
+        public string Name { get; init; }
+        public string? Url { get; init; }
+    }
+}
--- a/src/Sharpliner/GitHubActions/Job.cs
+++ b/src/Sharpliner/GitHubActions/Job.cs
@@ -1,7 +1,65 @@
+using System;
+using System.Collections.Generic;
+
 namespace Sharpliner.GitHubActions
 {
     public record Job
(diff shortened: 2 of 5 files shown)


src/Sharpliner/GitHubActions/Environment.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// <summary>
+    ///
+    /// </summary>
+    public record Environment
+    {
+        public Environment(string name, string? url = null)
+        {
+            if (string.IsNullOrEmpty(name))
+                throw new ArgumentNullException(nameof (name));
+            Name = name;
```
premun, line 14:
```suggestion
            Name = name ?? throw new ArgumentNullException(nameof(name));
```

src/Sharpliner/GitHubActions/Job.cs
```diff
@@ -1,7 +1,65 @@
...
+
 namespace Sharpliner.GitHubActions
 {
     public record Job
     {
+        public Job(string id)
+        {
+            if (string.IsNullOrEmpty(id))
+                throw new ArgumentNullException((nameof (id)));
+            Id = id;
```
premun, line 12:
```suggestion
            Id = id ?? throw new ArgumentNullException((nameof(id)));
```

premun (approved):

src/Sharpliner/GitHubActions/Environment.cs
mandel-macaque: But like that it can be empty :)

src/Sharpliner/GitHubActions/Job.cs
mandel-macaque: Same, it can be empty

premun: Sorry, gotcha!

premun: Wait

premun: What is the usecase?

premun: Just more strict validation?

PR #33, opened 2021-05-19
[Model] Add the possibility of defining single stage AzDO pipelines

premun:
Also use that for our pipeline here


premun:
Superceeded by #35

Issue #172, opened 2022-02-14
`RepositoryResource.Trigger` should be of type `Trigger`, not `PipelineTrigger`


ModernRonin:
What the title says. The consequence is one cannot do path filtering with triggers set up via resources|repositories.

[This link](https://github.com/microsoft/azure-pipelines-yaml/blob/master/design/pipeline-triggers.md#schema-1) documents the yaml behind it, but it's really the same as you have already implemented for `Trigger`.

PR #173, opened 2022-02-14
changed RepositoryResource.Trigger to be of type Trigger instead of P…

ModernRonin:
fix for  #172


premun (approved):

premun:
Thanks! I will have a new version for you shortly

PR #174, opened 2022-02-17
Fix PowerShell and Bash tasks when referencing script files

### LINKED ISSUE #168: PowershellFileTask generates invalid yaml
Hi,

maybe doing something wrong, but
```csharp
new PowershellFileTask("tools/scripts/RunDocker.ps1")
{
    Arguments = args,
    DisplayName = "Running docker image",
    Pwsh = true
};
```

creates this yaml:
```yaml
name: CreateReleasableImages

trigger:
  branches:
    include:
    - master
    - STABLE

variables:
- group: CommonPipelineVariables

- name: DOTNET_SKIP_FIRST_TIME_EXPERIENCE
  value: true

jobs:
- job: job
  pool:
    vmImage: ubuntu-latest
  steps:
  - checkout: self
    fetchDepth: 1

  - powershell: tools/scripts/RunDocker.ps1
    targetType: filepath
    arguments: -image ourcontainerregistry.azurecr.io/tools/pulumi:latest -command "/usr/bin/pwsh -Command {./vso.ps1; ./LogIntoAzure.ps1; ./RunPulumi.ps1 }" -doEnableDocker $true -$volumeMounts $(Build.SourcesDirectory)/image-maker:~/project
    displayName: Running docker image
    pwsh: true
```

which DevOps doesn't like:
>/tools/Pipelines/CreateReleasableImages.yml (Line: 31, Col: 5): Unexpected value 'targetType'
/tools/Pipelines/CreateReleasableImages.yml (Line: 32, Col: 5): Unexpected value 'arguments'
/tools/Pipelines/CreateReleasableImages.yml (Line: 34, Col: 5): Unexpected value 'pwsh'

Looking at the [documentation of the task](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/powershell?view=azure-devops), I get the impression that `powershell` is only for inline scripts and if one wants to use a file, one has to use the `PowerShell@2` task.

premun:
Fixes #168 

A little bit unfortunate solution but I found no other better. I wanted to preserve the nice `powershell: inline script` notation so I make an exemption when serializing the tasks that reference a script file so that they serialize as the full task invocations

### DIFF
--- a/src/Sharpliner/AzureDevOps/Model/Tasks/AzureDevOpsTask.cs
+++ b/src/Sharpliner/AzureDevOps/Model/Tasks/AzureDevOpsTask.cs
@@ -1,4 +1,5 @@
-﻿using Sharpliner.AzureDevOps.ConditionedExpressions;
+﻿using System;
+using Sharpliner.AzureDevOps.ConditionedExpressions;
 using YamlDotNet.Serialization;
 
 namespace Sharpliner.AzureDevOps.Tasks;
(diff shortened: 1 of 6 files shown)


premun:
cc @ModernRonin

PR #187, opened 2022-03-13
Add validation for repository checkouts

### LINKED ISSUE #164: Validate checkout repository references
Say, you do something like:

```csharp
 new RepositoryCheckoutTask("Cloud") { FetchDepth = 1 }
```

without having defined the repository under resources.

currently, it will compile without error only when trying to run the pipeline on Devops, you'll get a message
> Checkout of repository 'Cloud' is not supported. Only 'self', 'none', or a repository alias are supported.

It would be nice (and I think in the spirit of Sharpliner) if this could be caught at compile time.

Cheers,
Markus / MR

premun:
Resolves #164

### DIFF
--- a/src/Sharpliner/AzureDevOps/Pipeline.cs
+++ b/src/Sharpliner/AzureDevOps/Pipeline.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using Sharpliner.AzureDevOps.ConditionedExpressions;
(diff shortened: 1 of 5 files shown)


premun:
@ModernRonin this PR adds the validation for repository checkouts. I have a new validation plugin system in place that allows writing new plugins easily

ModernRonin:
super! sorry that I haven't gotten round to do anything much here lately, lots of release stress in my breadjob

premun:
That's quite alright! Can you validate later for me that it does what you intended? (either from a quick look at the PR or in your pipeline)

PR #199, opened 2022-05-13
Target .NET 6.0 in CodeQL only

premun:
CodeQL can't use .NET 7.0 so we need to do with this one only

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -40,7 +40,7 @@ jobs:
       shell: pwsh
       
     - name: Build
-      run: dotnet build src/Sharpliner/Sharpliner.csproj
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       shell: pwsh
       
     - name: Build
-      run: dotnet build src/Sharpliner/Sharpliner.csproj
+      run: dotnet build src/Sharpliner/Sharpliner.csproj -f net6.0 # CodeQL cannot do .NET 7.0 at the moment
```
premun, line 43:
```suggestion
      run: dotnet build src/Sharpliner/Sharpliner.csproj -f net6.0 /p:UseSharedCompilation=false # CodeQL cannot do .NET 7.0 at the moment
```

Issue #203, opened 2022-06-21
We should add implicit casts from string, i.e. to pool


premun:
It would be nice to have this:
```csharp
	Pool = "windows-latest"
```
as a shorthand for
```csharp
	Pool = new HostedPool(vmImage: "windows-latest")
```

AHeiming:
Hello,

I would like to work on this, if this is okay for you.

premun:
Absolutely! Thanks!

Please give it some thought, I wasn't especially verbose with the issue as it occurred to me briefly today.
I think vmImage makes the most sense but maybe we could also set a tuple and have the pool name too? Totally up to you, what you think is useful

premun:
Ah, sorry seems like homepage lost the "How to develop".

The CI project is a bit tricky and you can either unload it and work without it or you need to `dotnet build` the CI project from console.
You can then dotnet clean and build it in case you want to see the latest changes in the project.

What it does is build a package of the 43 version locally and the CI project loads it like it would get it from nuget.org to E2E test it. You can then also test the new feature there if you clean/build. You can also test it inside of the unit test project which has a direct reference.

Let me know if you have any other problems

AHeiming:
Hi thanks for the help, it works now.
Regarding the issue:

I think there might be some confusion if we implement this.
`Pool` is the base class of `HostedPool`. So every time a `Pool` is demanded, it does not necessarily has to be an instance of `HostedPool`.

My current approach was using the implicit operator to accomplish this.
But if we do this, everytime in the code a `Pool ` is instantiated without a constructor it will be of type `HostedPool`. Is this the desired behaivor?

I don't know how often `HostedPool` is chosen over the other Pools, I just want to avoid confusion.
This is my current approach:

`public static implicit operator Pool(string vmImage)
    {
        return new HostedPool(vmImage: vmImage);
    }`


I hope this was understandable 😄

premun:
Yes, I think that's roughly what I thought it should be. It's quite simple I guess but if I'm not wrong the YAML would then look almost the same and mean the same?

AHeiming:
Okay, I just wanted to be sure that this is the desired behaivour. 😄 
And yes, regarding the YAML file, I think so. 
I hope to be able to send a pull request today/tomorrow

premun:
Great! I'll be looking forward to it

premun:
Resolved by #209

PR #209, opened 2022-06-29
Fix Issue #203 (Implict cast from string to Pool)

AHeiming:
Issue #203 => Add implicit cast from string to Pool (resolves to HostedPool with vmImage).


premun:
Yeah, I guess that's it! Thanks!

In case you are interested in something more advanced, let me know, there's definitely something I can find you!

premun (approved):

PR #228, opened 2022-08-26
If.StartsWith and If.EndsWith

thomhurst:


premun (approved):

premun:
Thank you for the contribution!

Issue #227, opened 2022-08-26
Automatically wrap string expressions in condition in apostrophes


thomhurst:
Take this variable condition:

```csharp
        Variables =
        {
            StartsWith("refs/head/feature/", variables.Build.SourceBranch).Variable("myvariable", "pr-variable")
                .Else.Variable("myvariable", "other-variable"),
        },
```

The yaml being produced is:

```yaml
variables:
- ${{ if startsWith($(Build.SourceBranch), refs/head/feature/) }}:
  - name: myvariable
    value: pr-variable
- ${{ else }}:
  - name: myvariable
    value: other-variable
```

I get this exception:
![image](https://user-images.githubusercontent.com/30480171/186944149-ed1b5c6f-908c-406f-97d1-4cf3c7ffb8e4.png)

I think both of these values in the Contains/StartsWith/EndsWith conditions need to be made into strings. 
i.e. adding quotes around them

thomhurst:
I've proposed this solution. Let me know if you think it's bad.

#229

premun:
Thanks again for the contribution, @thomhurst! You can find the new updates in the `1.4.0` package.

Actually, with the `if does not start with '` it's probably not even a breaking change but I have already tagged it like `1.4.0` so we have that now.

PR #229, opened 2022-08-27
Wrap string condition expressions in apostrophes automatically

thomhurst:
StringCondition class for wrapping strings in quotes.

This class is to be inherited by String-type conditions such as Contains, StartsWith and EndsWith.

See issue #227

### DIFF
--- a/src/Sharpliner/AzureDevOps/ConditionedExpressions/Condition.cs
+++ b/src/Sharpliner/AzureDevOps/ConditionedExpressions/Condition.cs
@@ -71,7 +71,8 @@ public abstract class Condition
 
     protected static string WrapQuotes(string value)
     {
-        if (value.StartsWith(VariableIndexAccessStart)
+        if (value.StartsWith('\'')
+            || value.StartsWith(VariableIndexAccessStart)
             || value.StartsWith(VariablePropertyAccessStart))
         {
             return value;
@@ -146,7 +147,7 @@ public class NotCondition : Condition
     }
 }
 
-public class EqualityCondition : Condition
+public class EqualityCondition : StringCondition
 {
     internal EqualityCondition(bool equal, string expression1, string expression2)
         : base(equal ? "eq" : "ne", true, expression1, expression2)
@@ -291,13 +292,14 @@ public class ElseCondition<T> : Condition<T>
     public override string ToString() => ElseTagStart + ElseTagEnd;
 }
 
-public class EqualityCondition<T> : Condition<T>
+public class EqualityCondition<T> : StringCondition<T>
 {
     internal EqualityCondition(bool equal, string expression1, string expression2)
         : base(equal ? "eq" : "ne", true, expression1, expression2)
     {
     }
 }
+
 
 public class AndCondition<T> : Condition<T>
 {
@@ -401,6 +403,7 @@ public class LessCondition<T> : Condition<T>
     {
     }
 }
+
 
 public class BranchCondition : EqualityCondition
 {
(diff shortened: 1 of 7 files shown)


src/Sharpliner/AzureDevOps/ConditionedExpressions/Condition.cs
```diff
@@ -66,6 +69,17 @@ public static string RemoveTags(string condition)
         return condition;
     }
 
+    protected static string WrapQuotes(string value)
```
premun, line 72: Can you use this method in `BranchCondition` too?

tests/Sharpliner.Tests/AzureDevOps/ConditionedExpressions/ConditionalsTests.cs
```diff
@@ -271,4 +271,85 @@ public void Custom_Condition_Test()
...
+    {
+        public override Pipeline Pipeline => new()
+        {
+            Variables =
+            {
+                If.Contains("refs/heads/feature/", variables.Build.SourceBranch)
+                    .Variable("feature", "on"),
+
+                If.Contains("refs/heads/feature/", variables["Build.SourceBranch"])
+                .Variable("feature", "on")
```
premun, line 285:
```suggestion
                    .Variable("feature", "on")
```

premun (commented):
I think this is a nice improvement! I was trying to think of a case where we would force these and force the YAML to break with this but cannot think of any.

It's a breaking change, but hopefully that's okay. I will bump the minor version number.

Left just two minor comments, thanks!

src/Sharpliner/AzureDevOps/ConditionedExpressions/Condition.cs
thomhurst: BranchCondition inherits from EqualityCondition. I made that base class Inherit from String condition. 
What do you think?

It surfaced some existing tests that were asserting an incorrect syntax, so I've fixed those now too.

tests/Sharpliner.Tests/AzureDevOps/ConditionedExpressions/ConditionalsTests.cs
thomhurst: Fixed :)

src/Sharpliner/AzureDevOps/ConditionedExpressions/Condition.cs
premun: Great, that is the best outcome

tests/Sharpliner.Tests/AzureDevOps/ConditionedExpressions/ConditionalsTests.cs
```diff
@@ -31,7 +31,7 @@ public void And_Condition_Test()
         variable.Condition!.ToString().Should().Be(
             "and(" +
                 "notIn('bar', 'foo', 'xyz', 'foo'), " +
-                "ne($(Configuration), 'Debug'), " +
+                "ne('$(Configuration)', 'Debug'), " +
```
thomhurst, line 34: These tests were asserting an incorrect syntax. I made EqualityCondition inherit from StringCondition and it fixed this incorrect syntax, so I've updated all these tests to work now.

premun: I can see there are some failing around quoting `boolean` values. Do you know what's the correct syntax in YAML? Does it matter if you specify variable is `true` and compare with `'true'`?

premun: Thanks for trying it out!

premun (approved):

Issue #232, opened 2022-08-31
Variables don't serialize into Macro Expression syntax


thomhurst:
Given this C#

```csharp
Script.Inline($"echo {variables["MyVariable"]}"),
```

The yaml produced is:

```yaml
  - script: |-
      echo variables['MyVariable']
```

This doesn't work. For variables to work in string/variable interpolation they need to use Macro expression syntax.

This would be:
```yaml
  - script: |-
      echo $(MyVariable)
```

thomhurst:
#233

premun:
I see some nullability problems on the main build after the merge. I will do one more sanity check in the evening before releasing a new version

thomhurst:
Hope it's all good :)

premun:
A new version is out! 1.4.2

PR #233, opened 2022-08-31
Runtime + Compile + Macro Syntaxes for Variables and Parameters

thomhurst:
This provides strongly typed objects where users can call:

`Parameter.RuntimeExpression` => "parameters.Parameter"

`Parameter.CompileTimeExpression` => "{{ parameters.Parameter }}"

`Variable.MacroExpression` => "$(Variable)"

`Variable.RuntimeExpression` =>  "variables['Variable']"

`Variable.CompileTimeExpression` => "{{ variables.Variable }}

This gives the caller more control if they need it.

There's also new types with implicit conversions.

Strings implicitly convert to StringRuntimeExpression for string conditions.
StringRuntimeExpression is an IRuntimeExpression. This means conditions can accept these raw string expressions, or the new variable objects, or the new parameter objects. Then we can pull the correct run-time syntax out of these objects.

Variables now also serialize to Macro syntax so they can be used in scripts and such, but still serialize to runtime syntax for conditions and such.

This PR might seem quite convoluted but I couldn't think of a cleaner way to do this.
Let me know what you think.

See #232

### DIFF
--- a/src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
+++ b/src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
@@ -124,6 +124,24 @@ public class ConditionBuilder
     public Condition ContainsValue(IRuntimeExpression needle, params object[] haystack)
         => new ContainsValueCondition(needle, haystack.AsRuntimeExpressions());
 
+    public Condition In(StringRuntimeExpression needle, StringRuntimeExpression[] haystack)
+        => new InCondition(needle, haystack);
+
+    public Condition In(StringRuntimeExpression needle, IRuntimeExpression[] haystack)
+        => new InCondition(needle, haystack);
+
+    public Condition In(IRuntimeExpression needle, StringRuntimeExpression[] haystack)
+        => new InCondition(needle, haystack);
+
+    public Condition In(IRuntimeExpression needle, IRuntimeExpression[] haystack)
+        => new InCondition(needle, haystack);
+
+    public Condition In(StringRuntimeExpression needle, object[] haystack)
+        => new InCondition(needle, haystack.AsRuntimeExpressions());
+
+    public Condition In(IRuntimeExpression needle, object[] haystack)
+        => new InCondition(needle, haystack.AsRuntimeExpressions());
+
     public Condition In(StringRuntimeExpression needle, StringRuntimeExpression haystack)
         => new InCondition(needle, haystack);
 
@@ -153,6 +171,18 @@ public class ConditionBuilder
 
     public Condition NotIn(IRuntimeExpression needle, params object[] haystack)
         => new NotInCondition(needle, haystack.AsRuntimeExpressions());
+
+    public Condition NotIn(StringRuntimeExpression needle, StringRuntimeExpression haystack)
+        => new NotInCondition(needle, haystack);
+
+    public Condition NotIn(StringRuntimeExpression needle, IRuntimeExpression haystack)
+        => new NotInCondition(needle, haystack);
+
+    public Condition NotIn(IRuntimeExpression needle, StringRuntimeExpression haystack)
+        => new NotInCondition(needle, haystack);
+
+    public Condition NotIn(IRuntimeExpression needle, IRuntimeExpression haystack)
+        => new NotInCondition(needle, haystack);
 
     public Condition Greater(StringRuntimeExpression first, StringRuntimeExpression second)
         => Link(new GreaterCondition(first, second));
--- a/src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
+++ b/src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
@@ -1,4 +1,6 @@
-﻿namespace Sharpliner.AzureDevOps.ConditionedExpressions;
+﻿using Sharpliner.AzureDevOps.ConditionedExpressions.Interfaces;
+
+namespace Sharpliner.AzureDevOps.ConditionedExpressions;
 
 /// <summary>
 /// The builder is what let's us start the definition with the "If."
--- a/src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
+++ b/src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
@@ -118,6 +118,12 @@ public class ConditionBuilder
     public Condition ContainsValue(IRuntimeExpression needle, params IRuntimeExpression[] haystack)
         => new ContainsValueCondition(needle, haystack);
 
+    public Condition ContainsValue(StringRuntimeExpression needle, params object[] haystack)
+        => new ContainsValueCondition(needle, haystack.AsRuntimeExpressions());
+
+    public Condition ContainsValue(IRuntimeExpression needle, params object[] haystack)
+        => new ContainsValueCondition(needle, haystack.AsRuntimeExpressions());
+
     public Condition In(StringRuntimeExpression needle, StringRuntimeExpression haystack)
         => new InCondition(needle, haystack);
 
@@ -141,6 +147,12 @@ public class ConditionBuilder
 
     public Condition NotIn(IRuntimeExpression needle, params IRuntimeExpression[] haystack)
         => new NotInCondition(needle, haystack);
+
+    public Condition NotIn(StringRuntimeExpression needle, params object[] haystack)
+        => new NotInCondition(needle, haystack.AsRuntimeExpressions());
+
+    public Condition NotIn(IRuntimeExpression needle, params object[] haystack)
+        => new NotInCondition(needle, haystack.AsRuntimeExpressions());
 
     public Condition Greater(StringRuntimeExpression first, StringRuntimeExpression second)
         => Link(new GreaterCondition(first, second));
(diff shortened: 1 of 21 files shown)


premun (changes requested):
I don't think this is convoluted at all. I think it's a really clever use of implicit conversions and I think this might be a priceless feature for Sharpliner!

However, I will need to do more testing to check for myself how it works with the 4 different combinations of the methods we need and whether the compiler can always figure things out.

In the meantime, I left some minor comments mostly around styling but there is one situation where this new unboxing does not work but I believe it can be fixed.

If you open `TemplateTests` there are some expressions in the very last test (`Conditioned_Parameters_Serialization_Test`). If you modify the pipeline used there and use the `parameters["container"]` strong typed notation, you will see the problem.
The problem is that we are creating there a `TemplateParameter` dictionary which is a dictionary that supports conditioned expressions. In this case, we are not implicitly casting into `string` because `TemplateParameter` expect an `object`. Then, the YAML serializer just serializes the into the `ParameterValue` object directly ending up with this:

```yaml
    parameterName: container
    runtimeExpression: parameters.container
    compileTimeExpression: ${{ parameters.container }}"
```

Now I _think_ you can solve this by making `ParameterValue` implement `IYamlConvertible` and implement the `Write` to serialize into `${{ parameters.container }}` if I'm not mistaken. But maybe it has another solution that I can't think of now.

Can you have a look at that?

src/Sharpliner/AzureDevOps/ConditionedExpressions/Interfaces/ICompileTimeExpression.cs
```diff
@@ -0,0 +1,6 @@
+namespace Sharpliner.AzureDevOps.ConditionedExpressions.Interfaces;
+
+public interface ICompileTimeExpression
```
premun, line 3: Could you add a short XML doc to each of these and what they represent? It might be helpful for future

tests/Sharpliner.Tests/AzureDevOps/StrongTypedTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+        };
+    }
+
+    [Fact]
+    public void Variable_And_Parameter_Serialize_Differently_Within_Scripts_And_Conditions_Test()
+    {
+        var yaml = new Pipeline_With_Strong_Variables_And_Parameters().Serialize();
+
+        yaml.Trim().Should().Be(@"parameters:
```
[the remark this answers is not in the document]
thomhurst: Done

src/Sharpliner/AzureDevOps/ConditionedExpressions/VariableValue.cs
```diff
@@ -0,0 +1,23 @@
...
+    public string VariableName { get; }
+
+    internal VariableValue(string variableName)
+    {
+        VariableName = variableName;
+    }
+
+    public string RuntimeExpression => $"variables['{VariableName}']";
+
+    public string CompileTimeExpression => $"${{{{ variables.{VariableName} }}}}";
```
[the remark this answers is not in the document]
thomhurst: Done

src/Sharpliner/AzureDevOps/ConditionedExpressions/ParameterValue.cs
```diff
@@ -0,0 +1,21 @@
...
+    public string ParameterName { get; }
+
+    internal ParameterValue(string parameterName)
+    {
+        ParameterName = parameterName;
+    }
+
+    public string RuntimeExpression => $"parameters.{ParameterName}";
+
+    public string CompileTimeExpression => $"${{{{ parameters.{ParameterName} }}}}";
```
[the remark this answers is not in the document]
thomhurst: Done

src/Sharpliner/AzureDevOps/ConditionedExpressions/ParameterValue.cs
```diff
@@ -0,0 +1,21 @@
+using Sharpliner.AzureDevOps.ConditionedExpressions.Interfaces;
+
+namespace Sharpliner.AzureDevOps.ConditionedExpressions;
+
+public class ParameterValue : IRuntimeExpression, ICompileTimeExpression
```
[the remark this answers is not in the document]
thomhurst: Done

src/Sharpliner/AzureDevOps/ConditionedExpressions/Interfaces/ICompileTimeExpression.cs
thomhurst: Done

thomhurst:
The IYamlConvertible worked !

src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
```diff
@@ -129,13 +200,40 @@ public Condition<T> And(params string[] expressions)
...
+    public Condition<T> Contains(IRuntimeExpression needle, StringRuntimeExpression haystack)
+        => new ContainsCondition<T>(needle, haystack);
+
+    public Condition<T> Contains(StringRuntimeExpression needle, IRuntimeExpression haystack)
+        => new ContainsCondition<T>(needle, haystack);
+
+    public Condition<T> Contains(IRuntimeExpression needle, IRuntimeExpression haystack)
         => new ContainsCondition<T>(needle, haystack);
 
     public Condition In(string needle, string haystack)
```
premun, line 239: So this is my fault, not yours, but we're missing some APIs here. Would you mind completing those too? We can also do it as part of another PR but maybe it makes sense?

Something like 
```csharp
    public Condition In(string needle, string haystack)
        => new InCondition(needle, haystack);

    public Condition In(string needle, params string[] haystack)
        => new InCondition(needle, haystack);

    public Condition NotIn(string needle, string haystack)
        => new NotInCondition(needle, haystack);

    public Condition NotIn(string needle, params string[] haystack)
        => new NotInCondition(needle, haystack);
```
(not sure these might not need changing to `IRuntimeExpression` then.

What should be possible is something like this:
```csharp
Variables =
{
    If.In(variables.Build.SourceBranch, "refs/heads/feature/", parameters["allowedTags"], variables["foo"], variables.Build.Reason)
        .Variable("feature", "on"),
}
```

src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
```diff
@@ -147,10 +245,16 @@ public Condition NotIn(string needle, params string[] haystack)
...
+    public Condition<T> IsBranch(IRuntimeExpression branchName)
+        => Link(new BranchCondition<T>(branchName, true));
+
+    public Condition<T> IsNotBranch(StringRuntimeExpression branchName)
+        => Link(new BranchCondition<T>(branchName, false));
+
+    public Condition<T> IsNotBranch(IRuntimeExpression branchName)
         => Link(new BranchCondition<T>(branchName, false));
 
     public Condition<T> Greater(string first, string second)
```
premun, line 260: `Greater` and `Less` should ideally also work with the new expressions

src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
```diff
@@ -105,10 +158,28 @@ internal ConditionBuilder(Conditioned<T>? parent = null)
...
+    public Condition<T> NotEqual(IRuntimeExpression expression1, StringRuntimeExpression expression2)
+        => Link(new EqualityCondition<T>(false, expression1, expression2));
+
+    public Condition<T> NotEqual(StringRuntimeExpression expression1, IRuntimeExpression expression2)
+        => Link(new EqualityCondition<T>(false, expression1, expression2));
+
+    public Condition<T> NotEqual(IRuntimeExpression expression1, IRuntimeExpression expression2)
         => Link(new EqualityCondition<T>(false, expression1, expression2));
 
     public Condition<T> And(params Condition[] expressions)
```
premun, line 185: `And` and `Or` in my opinion should not support the new stuff, so I'd keep them as-is

premun (commented):
Looks good!

I found couple of more missing APIs but from my side and already missing in Sharpliner. Please let me know if you are rushing with having this PR in and we can merge right away and add the missing ones as a follow-up, or whether you could also add them as part of this PR.

src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
thomhurst: Done

src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
thomhurst: Done

src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionBuilder.cs
thomhurst: I had to add two extra methods to accept a params object[] array. This is so the caller can mix variables/params (IRuntimeExpression) and strings within the same array.

But this should be done now if you'd like to take a look. Added a test for your above scenario too.

premun:
@thomhurst this is really amazing, thank you!

I have actually heard many times online from people that mixing these types of variables causes them a lot of headache so I will update the README later to list this as one of the main features.

premun (approved):

PR #235, opened 2022-09-07
Change built-in variables to strong-typed objects

thomhurst:
Change built-in variables to strong-typed objects to benefit from the new Macro / Runtime syntaxes


thomhurst:
Yeah I was trying to do a condition if source branch starts with something and it wasn't working. It was trying to do that macro syntax before. I manually referenced the variable by name to get a variable strong type, which then serialised into the runtime syntax and my condition started working.

premun:
Right, that makes sense. There should be version `1.4.3` on NuGet in a couple of minutes with this fix

PR #237, opened 2022-09-16
Include default values for parameters when one is set

### LINKED ISSUE #236: Parameter Default Values are not used if they're value type defaults
I've tried declaring a BooleanParameter with a default value of false, but this doesn't serialise into the pipeline.

I then get an error when running the pipeline saying I must set a parameter value (since it has no default)

I suspect the same would be if you tried an integer with a value of 0.

premun:
Resolves #236


premun:
cc @thomhurst

PR #240, opened 2022-09-23
Check all public API changes with a test

premun:
This adds a tests that validates whether there are no unwanted public API changes in the main library


premun:
@thomhurst I am adding a test here that compares changes in the public API of the library. It will be required so that we have a history of what changed to catch for unwanted removals/changes and validate new additions to the public surface of the library. I will merge this now and you will have to merge/rebase your PR to accommodate for this.

Issue #259, opened 2022-12-29
ObjectParameter but with List instead of Dictionary?


mathiasi:
I'm testing out Sharpliner and trying to migrate our current pipeline.
We have one parameter as following:

```
- name: myName
  displayName: My List
  type: object
  default: ['value 1', 'value 2']
```
and I can't seem to replicate this in Sharpliner. As far as I can tell there is only one ObjectParameter and it only accepts a Dictionary.

I was trying to implement a `Parameter<ConditionedList<string>>` but the constructors of Parameter is internal. Would it be worth it to consider making it public? I might have been able make my own implementation by doing so.

premun:
Support added in `1.4.9` - https://github.com/sharpliner/sharpliner/releases/tag/1.4.9

premun:
Thanks again, @mathiasi

PR #260, opened 2022-12-29
Added support list object parameter

mathiasi:
Attempt to add support for #259


premun:
Hi @mathiasi,

Thanks for the contribution! I will shortly review Azure Pipeline docs to check how they name this and review in a bit. If all is well, we can have a new version out in about an hour or so.

premun:
Yeah, this makes sense! Thanks again

premun (approved):

PR #261, opened 2023-01-27
Fixed wrong property being outputted for TestRunTitle in DotNetTestCoreCliTask

mathiasi:


premun (approved):

premun:
Hi @mathiasi, thanks for the change! I will merge it first thing tomorrow and publish a new version of the library

premun:
@mathiasi there will be the version `1.4.10` containing the fix in a couple of minutes available on nuget.org. Thanks again for the contribution

PR #263, opened 2023-01-31
Support for LockBehavior

thomhurst:
Support for LockBehavior

### DIFF
--- a/src/Sharpliner/AzureDevOps/LockBehaviour.cs
+++ b/src/Sharpliner/AzureDevOps/LockBehaviour.cs
@@ -1,10 +1,12 @@
-﻿using System.Diagnostics.CodeAnalysis;
+﻿using YamlDotNet.Serialization;
 
 namespace Sharpliner.AzureDevOps;
 
-[SuppressMessage("ReSharper", "InconsistentNaming")]
 public enum LockBehaviour
 {
-    runLatest,
-    sequential
+    [YamlMember(Alias = "runLatest")]
+    RunLatest,
+
+    [YamlMember(Alias = "sequential")]
+    Sequential
 }
--- a/src/Sharpliner/AzureDevOps/LockBehaviour.cs
+++ b/src/Sharpliner/AzureDevOps/LockBehaviour.cs
@@ -1,0 +1,10 @@
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace Sharpliner.AzureDevOps;
+
+[SuppressMessage("ReSharper", "InconsistentNaming")]
+public enum LockBehaviour
+{
+    runLatest,
+    sequential
+}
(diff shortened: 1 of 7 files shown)


src/Sharpliner/AzureDevOps/LockBehaviour.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace Sharpliner.AzureDevOps;
+
+[SuppressMessage("ReSharper", "InconsistentNaming")]
+public enum LockBehaviour
+{
+    runLatest,
+    sequential
```
premun, line 9: In other parts of the library, we stayed more C#-like and kept naming capitalized. I am wondering whether that would not be better.

But I understand it's complicated here as this is wrapped in the `Conditioned<>` later.

Can I ask you to add a following serialization helper?

```csharp
public class YamlStringEnumConverter : IYamlTypeConverter
{
    public bool Accepts(Type type) => type.IsEnum;

    public object ReadYaml(IParser parser, Type type) => throw new NotImplementedException();

    public void WriteYaml(IEmitter emitter, object? value, Type type)
    {
        if (value is null)
        {
            emitter.Emit(new Scalar(null!));
            return;
        }

        var enumMember = type.GetMember(value.ToString()!).FirstOrDefault();
        var yamlValue = enumMember?
            .GetCustomAttributes<YamlMemberAttribute>(true)
                .Select(ema => ema.Alias)
                .FirstOrDefault()
            ?? value.ToString();
        
        emitter.Emit(new Scalar(yamlValue!));
    }
}
```

You can do following then:
```csharp
public enum LockBehaviour
{
    [YamlMember(Alias = "runLatest")]
    RunLatest,

    [YamlMember(Alias = "sequential")]
    Sequential
}
```

premun: I will then later apply to other places in the code that could use this.

premun: Oh, and register it as well:

```
    private static ISerializer InitializeSerializer()
    {
        var serializerBuilder = new SerializerBuilder()
            .WithNamingConvention(CamelCaseNamingConvention.Instance)
            .WithTypeConverter(new YamlStringEnumConverter())      <<<<<<<<<<<<<<<
            .ConfigureDefaultValuesHandling(DefaultValuesHandling.OmitDefaults | DefaultValuesHandling.OmitEmptyCollections)
            .WithEventEmitter(nextEmitter => new MultilineStringEmitter(nextEmitter));

        return serializerBuilder.Build();
    }
```

thomhurst: Nice one - Looks all good, tests seem happy

premun (approved):

Issue #296, opened 2024-04-08
Add models for PublishTestResults@2 and PublishCodeCoverageResults@1 ADO tasks


aditnryn:
### Issue
These tasks are widely used in CI/CD pipelines to publish test and coverage details. Today, defining them in CSharpier is not an optimal dev experience since it requires writing inputs by hand.

See https://github.com/sharpliner/sharpliner/blob/9cc9a805d8bbcf5c33a43dba7e2b30cce44c0c09/tests/Sharpliner.Tests/AzureDevOps/Pipelines/XHarnessPipeline.cs#L101C1-L113C1

### Goal
Add models for them (likely) in the https://github.com/sharpliner/sharpliner/tree/main/src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace directory.

PR #299, opened 2024-04-10
Add model for PublishTestResults@2

aditnryn:
Partial fix for #296

### DIFF
--- a/src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/PublishTestResultsTask.cs
+++ b/src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/PublishTestResultsTask.cs
@@ -1,0 +1,161 @@
+﻿using System;
+using Sharpliner.AzureDevOps.Tasks;
+using YamlDotNet.Serialization;
+
+namespace Sharpliner.AzureDevOps.Model.Tasks;
+
+/// <summary>
+/// More details can be found in <see href="https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/publish-test-results-v2">official Azure DevOps pipelines documentation</see>
+/// </summary>
+public record PublishTestResultsTask : AzureDevOpsTask
+{
+    public PublishTestResultsTask(TestResultsFormat testResultsFormat, string testResultsFile) : base("PublishTestResults@2")
+    {
+        TestResultsFormat = testResultsFormat;
+        TestResultsFiles = testResultsFile;
+    }
+
+    /// <summary>
+    /// Specifies the format of the results files you want to publish.
+    /// JUnit, NUnit, VSTest, XUnit, CTest
+    /// Defaults to <code>JUnit</code>.
+    /// </summary>
+    [YamlIgnore]
+    public TestResultsFormat TestResultsFormat
+    {
+        get => GetString("testResultsFormat") switch
+        {
+            "JUnit" => TestResultsFormat.JUnit,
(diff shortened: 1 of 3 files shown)


src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/PublishTestResultsTask.cs
```diff
@@ -0,0 +1,143 @@
...
+    /// <summary>
+    /// Specifies one or more test results files.
+    /// You can use a single-folder wildcard (*) and recursive wildcards (**).
+    /// <example>**/TEST-*.xml searches for all the XML files whose names start with TEST- in all subdirectories.</example>
+    /// <remarks>If using VSTest as the test result format, the file type should be changed to .trx e.g. **/TEST-*.trx</remarks>
+    /// </summary>
+    [YamlIgnore]
+    public string TestResultsFiles
+    {
+        get => GetString("testResultsFiles") ?? "**/TEST-*.xml";
```
premun, line 37: Why this default value?

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/PublishTestResultsTask.cs
```diff
@@ -0,0 +1,143 @@
...
+        init => SetProperty("testResultsFiles", value);
+    }
+
+    /// <summary>
+    /// Specifies the folder to search for the test result files.
+    /// </summary>
+    [YamlIgnore]
+    public string SearchFolder
+    {
+        get => GetString("searchFolder") ?? "$(System.DefaultWorkingDirectory)";
```
premun, line 47: Same - why do we have a default? Shouldn't we return `null`?

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/PublishTestResultsTask.cs
aditnryn: That's the default value when the input is not set as per docs

```
    testResultsFiles: '**/TEST-*.xml' # string. Required. Test results files. Default: **/TEST-*.xml.
```

I added it as a default so that if the property isn't set, the following `get`s return the value ADO considers as default.

Do you propose that I should get rid of the default altogether?

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/PublishTestResultsTask.cs
aditnryn: That's the default value when the input is not set as per docs

```
    #searchFolder: '$(System.DefaultWorkingDirectory)' # string. Search folder. Default: $(System.DefaultWorkingDirectory).
```

I added it as a default so that if the property isn't set, the following `get`s return the value ADO considers as default.

Do you propose that I should get rid of the default altogether?

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/PublishTestResultsTask.cs
aditnryn: Done

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/PublishTestResultsTask.cs
aditnryn: Removed defaults

premun (approved):

PR #305, opened 2024-04-16
Add an easy to way to export the public API

### LINKED ISSUE #303: Make it easier to export new API
It's quite an annoying process when you need to update the public API export through the unit test. We should have a script for this.

premun:
Resolves #303


premun:
@aditnryn you can now F5 the `PublicApiExporterProject` and it will regenerate the public API.

PR #317, opened 2024-07-23
Update Task template to add a task retry property

PaulieC1972:
An option was added to ADO tasks in 2021 to enable immediate retry of the task on failure. This allows for failures due to intermittency or network latency etc. to be retried without a user having to re-run the job from the UI.  This change adds the option of specifying retryCountOnTaskFailure as a string value, as specified here https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-task?view=azure-pipelines


PaulieC1972:
The output looks like this:
`- task: UseDotNet@2
  displayName: Install .NET Core SDK 6.0.x
  inputs:
    version: 6.0.x
    packageType: ${{ parameters.packageType }}
    installationPath: ${{ parameters.installationPath }}
  condition: ne('${{ parameters.skipUseDotNetTask }}', true)
  retryCountOnTaskFailure: 3
`

src/Sharpliner/AzureDevOps/Model/Step.cs
```diff
@@ -70,6 +70,13 @@ public abstract record Step
     [DisallowNull]
     public ConditionedDictionary Env { get; init; } = [];
 
+    /// <summary>
+    /// Number of retries if the task fails.
+    /// Default is 0
+    /// </summary>
+    [YamlMember(Order = 230)]
+    public Conditioned<string>? RetryCountOnTaskFailure { get; init; }
```
premun, line 78: Why not `int`?

```suggestion
    public Conditioned<int>? RetryCountOnTaskFailure { get; init; }
```

premun (commented):
Hi @PaulieC1972,
Thanks for the contribution!

For this build to go green, there is failing test with instructions that updates the public API of the library:

https://github.com/sharpliner/sharpliner/blob/b0363e435b1fb26b470a86f48d05082e1c509b06/tests/Sharpliner.Tests/PublicApiChangeTest.cs#L18-L35

src/Sharpliner/AzureDevOps/Model/Step.cs
PaulieC1972: I am perfectly fine with having an int, in fact I would prefer it as it saves typing quotes, but the docs specific as a string https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-task?view=azure-pipelines

PaulieC1972: I am perfectly fine with having an int, in fact I would prefer it as it saves typing quotes, but the docs specify it as a string https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/steps-task?view=azure-pipelines

premun: Nice catch but I'd just say the docs are off (that was my experience in many places) because I see `string` in `timeoutInMinutes` too. So let's just make this an `int` for a better experience because we can well expect people will be putting in numbers

premun, line 78: @PaulieC1972 I was told that this won't work on any `step` and maybe we should only put this on `task` to `AzureDevOpsTask`

premun: Seems to match the documentation which is for `step.task`

PaulieC1972: Yes, I thought that might be the case.  I'll look into this again, but I couldn't find a good place to add this in for task only

PaulieC1972: But completely understand that this retry is task bound and should not be available for steps

PaulieC1972:
@premun Is there an automated way to update PublicApiExport.txt ? I have added an entry, but test is still failing

premun:
> @premun Is there an automated way to update PublicApiExport.txt ? I have added an entry, but test is still failing

Yes, there is. Weird that it's not in the test exception.
```
cd eng/PublicApiExporter
dotnet run
```

PaulieC1972:
> > @premun Is there an automated way to update PublicApiExport.txt ? I have added an entry, but test is still failing
> 
> @PaulieC1972 yes, there is. Weird that it's not in the test exception.
> 
> ```
> cd eng/PublicApiExporter
> dotnet run
> ```

OK, I generated that locally and pushed the change but the PR test is still failing, so I am asuming that something else needs to be changed alongside ?

premun (approved):

premun:
I finished the PR - you will find the change in the package `1.5.19`

PR #323, opened 2024-10-27
Add the fetchFilter property to the checkout task

Meir017:


premun (approved):
Thanks for this contribution! I will publish a new version of the library with this change shortly

premun:
Oops, seems like my AzDO account is disabled

Issue #325, opened 2024-10-27
Document all public API's - Remove CS1591 NoWarn


Meir017:
All public API's should be documented :)

[build.log](https://github.com/user-attachments/files/17535441/build.log)

- [x] YamlStringEnumConverter - made internal
- [x] MultilineStringEmitter - should this be internal?
- [x] IDependsOn - pubic API 
- [x] SharplinerSerializer - made internal
- [x] TargetPathType
- [x] SharplinerConfiguration
- [x] TaskBuilderBase
- [x] VariablesReference #326
- [x] TemplateDefinitionCollection, TemplateDefinition, Template
- [x] PublicDefinitions
- [x] Pipeline, PipelineDefinition, PipelineDefinitionCollection
- [x] Parameter
- [x] VariableDefinition
- [x] Conditioned #334
- [x] ValidationsExtensions - public API
- [x] ISharplinerDefinitionCollection
- [x] IDefinitionValidation

tasks:
- [x] src\Sharpliner\AzureDevOps\Model\Tasks\AzureDevOpsTask - #352
- [x] src/Sharpliner/AzureDevOps/Model/Tasks/Builders/ScriptTaskBuilder.cs - #350 
- [x] src/Sharpliner/AzureDevOps/Model/Tasks/ScriptTask.cs - #350 
- [x] src\Sharpliner\AzureDevOps\Model\Tasks\BashTask - #358 
- [x] src\Sharpliner\AzureDevOps\Model\Tasks\Builders\BashTaskBuilder - #358
- [x] src\Sharpliner\AzureDevOps\Model\Tasks\Builders\CheckoutTaskBuilder - #357 
- [x] src\Sharpliner\AzureDevOps\Model\Tasks\CheckoutTask - #357 
- [x] src\Sharpliner\AzureDevOps\Model\Tasks\DownloadTask - #359
- [x] src\Sharpliner\AzureDevOps\Model\Tasks\Builders\DownloadTaskBuilder - #359
- [x] src\Sharpliner\AzureDevOps\Model\Tasks\Builders\PowerShellTaskBuilder - #360
- [x] src\Sharpliner\AzureDevOps\Model\Tasks\Builders\PwshTaskBuilder - #360
- [x] src\Sharpliner\AzureDevOps\Model\Tasks\PowerShellTask - #360
- [x] src\Sharpliner\AzureDevOps\Model\Tasks\Marketplace\ArchiveFilesTask - #362
[...]

PR #327, opened 2024-10-28
chore: don't expose the IYamlConvertible methods as public

Meir017:
This helps with #325 as these methods appear as public-API


premun (approved):
Thanks yet again. Reducing the API surface is a good idea

premun:
I need to figure out why the builds don't trigger

PR #335, opened 2024-10-28
Fix PublicApiExport.txt

Meir017:


premun:
It worked for quite some time and now it stopped again..

PR #340, opened 2024-10-30
feature: add Parameter.Reference to allow strongly-typed parameters

Meir017:
this will allow us to reference parameters in a strongly-typed way

I think something similar can be done for variables too

### DIFF
--- a/src/Sharpliner/AzureDevOps/Parameter.cs
+++ b/src/Sharpliner/AzureDevOps/Parameter.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using Sharpliner.AzureDevOps.ConditionedExpressions;
+using Sharpliner.AzureDevOps.ConditionedExpressions.Arguments;
 using YamlDotNet.Core;
 using YamlDotNet.Core.Events;
 using YamlDotNet.Serialization;
@@ -69,6 +70,18 @@ public abstract record Parameter
     /// </summary>
     [YamlIgnore]
     public ParameterReference Reference => new(Name);
+
+    /// <summary>
+    /// Converts a <see cref="Parameter"/> to a <see cref="IfExpression"/> by getting the reference to the parameter.
+    /// </summary>
+    /// <param name="parameter">The parameter.</param>
+    public static implicit operator IfExpression(Parameter parameter) => parameter.Reference;
+
+    /// <summary>
+    /// Converts a <see cref="Parameter"/> to a <see cref="InlineExpression"/> by getting the reference to the parameter.
+    /// </summary>
+    /// <param name="parameter">The parameter.</param>
+    public static implicit operator InlineExpression(Parameter parameter) => parameter.Reference;
 }
 
 /// <summary>
--- a/src/Sharpliner/AzureDevOps/Parameter.cs
+++ b/src/Sharpliner/AzureDevOps/Parameter.cs
@@ -9,7 +9,41 @@ using YamlDotNet.Serialization;
 namespace Sharpliner.AzureDevOps;
 
 /// <summary>
+/// <para>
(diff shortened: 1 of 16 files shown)


premun (approved):
This one is super nice! Could you also please add it to the [definition reference](https://github.com/sharpliner/sharpliner/blob/main/docs/AzureDevOps/DefinitionReference.md)?

Meir017:
@premun Done! added new overloads for the `*ParameterReference` methods to support this flow

premun:
@Meir017 I just had another idea - why don't we just add implicit conversion of `Parameter` to `string` or even one of the `Expression` objects so that we don't need to use the `Reference` field and can just plug the `Parameter` objects directly into the APIs that accept `string` for example

tests/Sharpliner.Tests/AzureDevOps/TemplateTests.cs
```diff
@@ -46,37 +46,44 @@ public void TemplateList_Serialization_Test()
...
+            skipBuild,
+            useNugetOrg,
+            restore,
+            afterBuild,
         ];
 
         public override ConditionedList<Step> Definition =>
         [
-            DotNet.Install.Sdk(parameters["version"]),
+            DotNet.Install.Sdk(version.Reference),
```
Meir017, line 72: in these cases what used to happen in that the `ParameterReference` type has an implicit operator to `string` so this worked.

I'm not sure about adding an implicit operator for string for the `Parameter` type will be clear, \
@premun  WDYT?

premun: I think it might just work because wherever you use this, you'd actually expect the `${{ parameter.name }}` result and if you would have this:
```
If.Equal(version, "8.0")
```

then it would do `${{ if eq(${{ parameter.version }}, "8.0" }} but I believe we already can recognize this and strip the inner `${{` away so it might just all work

premun: So maybe just `Parameter` -> `string` implicit conversion is all we need

premun: But I understand it's a dangerous thing that could happen invisibly in many places around your codebase and it can be dangerous

premun: I think the `.Reference` member is a good idea over all too, just to be clear. I am just thinking that maybe it won't be as discoverable and the auto-cast to string might just work out nice. After all, you don't handle the `Parameter` objects a lot and they do not get passed around much, so it might be just the right and nice thing to do.

src/Sharpliner/AzureDevOps/Parameter.cs
```diff
@@ -40,6 +41,24 @@ public Parameter(string name, string? displayName = null)
...
+    /// Converts a <see cref="Parameter"/> to a <see cref="IfExpression"/> by getting the reference to the parameter.
+    /// </summary>
+    /// <param name="parameter">The parameter.</param>
+    public static implicit operator IfExpression(Parameter parameter) => new ParameterReference(parameter.Name);
+
+    /// <summary>
+    /// Converts a <see cref="Parameter"/> to a <see cref="InlineExpression"/> by getting the reference to the parameter.
+    /// </summary>
+    /// <param name="parameter">The parameter.</param>
+    public static implicit operator InlineExpression(Parameter parameter) => new ParameterReference(parameter.Name);
```
Meir017, line 61: I think we need all 3 overloads since an operator for `ParameterReference` won't be enough for an implicit operator (but converting to these different types by leveraging the implicit operator of `ParameterReference` does the job here 😀)

premun (approved):
Nice, thanks for this!

PR #345, opened 2024-11-02
Update .NET SDK and resolve some compiler messages

premun:
With upgrades of the .NET SDK, new compiler messages surface so this is resolving some

### DIFF
--- a/eng/Sharpliner.CI/ProjectBuildSteps.cs
+++ b/eng/Sharpliner.CI/ProjectBuildSteps.cs
@@ -12,7 +12,7 @@ class ProjectBuildSteps(string project) : StepLibrary
     [
         StepTemplate(Pipelines.TemplateLocation + "install-dotnet-sdk.yml", new()
         {
-            { "version", "8.0.204" }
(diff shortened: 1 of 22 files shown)


eng/Sharpliner.CI/ProjectBuildSteps.cs
```diff
@@ -12,7 +12,7 @@ class ProjectBuildSteps(string project) : StepLibrary
     [
         StepTemplate(Pipelines.TemplateLocation + "install-dotnet-sdk.yml", new()
         {
-            { "version", "8.0.204" }
+            { "version", "8.0.403" }
```
Meir017, line 15: @premun would be nice to have a utility to expose a `GlobalJson` type that we could leverage for pipelines (this type would be static and read value for the local `globa.json` file

premun: Yeah, that sonuds like a good idea. I wonder if we should make that a feature of the library or just an internal feature for this repo

PR #347, opened 2024-11-03
devops: msbuild code-coverage allowing for XML coverage to be viewed in azure-pipelines

Meir017:
improving on #329

now locally the coverage is using https://cobertura.github.io/cobertura/ which doesn't depend on visual-studio so locally we can use https://github.com/danielpalme/ReportGenerator to generate an HTML report


premun (approved):
I like this. I am usually not overly tracking this but for the models we have, it looks like it will be actually very useful to see for which models we test serialization

PR #349, opened 2024-11-03
docs: complete core API's

Meir017:
related to #325

down to ~120 warnings, only for the task types


src/Sharpliner/AzureDevOps/VariablesReference.cs
```diff
@@ -60,8 +60,6 @@ public class VariablesReference
     /// This variable is agent-scoped, and can be used as an environment variable in a script and as a parameter in a build task, but not as part of the build number or as a version control tag.
     /// </summary>
     public VariableReference TF_BUILD => new("TF_BUILD");
-
-    public VariableReference Configuration => new("Configuration");
```
Meir017, line 64: this doesn't exist in https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables

premun (approved):

premun:
> related to #325
> 
> down to ~120 warnings, only for the task types

Oh wow, a lot of progress!

PR #351, opened 2024-11-03
chore: fix azure-pipelines link

Meir017:
This build id is not 6, not 4


premun (approved):
Ah, yeah I recreated it when the pipelines were not triggering the other day

PR #366, opened 2024-11-11
feature: Variable as reference

Meir017:
closes #354

PR #367, opened 2024-11-11
docs: variable as reference

Meir017:
continuation of #366

### DIFF
--- a/docs/AzureDevOps/GettingStarted.md
+++ b/docs/AzureDevOps/GettingStarted.md
@@ -79,7 +79,7 @@ An example of a pipeline that builds and tests your PR can look like this:
 ```csharp
 ...
 
-private static readonly Variable s_DotnetVersionVariable = new Variable("DotnetVersion", string.Empty);
+private static readonly Variable DotnetVersion = new Variable("DotnetVersion", string.Empty);
 
 public override SingleStagePipelineDefinition Pipeline => new()
 {
@@ -88,10 +88,10 @@ public override SingleStagePipelineDefinition Pipeline => new()
     Variables =
     [
         If.IsBranch("net-6.0")
-            .Variable(s_DotnetVersionVariable with { Value = "6.0.100" })
+            .Variable(DotnetVersion with { Value = "6.0.100" })
             .Group("net6-kv")
         .Else
-            .Variable(s_DotnetVersionVariable with { Value = "5.0.202" }),
+            .Variable(DotnetVersion with { Value = "5.0.202" }),
     ],
 
     Jobs =
@@ -105,7 +105,7 @@ public override SingleStagePipelineDefinition Pipeline => new()
                     .Step(Powershell.Inline("Write-Host 'Hello-World'").DisplayAs("Hello world")),
 
                 DotNet.Install
-                    .Sdk(s_DotnetVersionVariable)
+                    .Sdk(DotnetVersion)
                     .DisplayAs("Install .NET SDK"),
 
                 DotNet
--- a/docs/AzureDevOps/GettingStarted.md
+++ b/docs/AzureDevOps/GettingStarted.md
@@ -59,7 +59,7 @@ class PullRequestPipeline : PipelineDefinition
 }
 ```
 
-You can also decide to override `SingleStagePipelineDefinition` in case you have a simpler pipeline with only a single stage.  
+You can also decide to override `SingleStagePipelineDefinition` in case you have a simpler pipeline with only a single stage.
 You can also overried `PipelineWithExtends` to [extend an existing template](https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/extends?view=azure-pipelines).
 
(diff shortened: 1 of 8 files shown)


Meir017:
test `Sharpliner.Tests.AzureDevOps.StrongTypedTests.Variable_And_Parameter_Serialize_Differently_Within_Scripts_And_Conditions_Test` fails

expected:
```yaml
parameters:
- name: Parameter1
  type: string
  default: SomeParameterValue1

- name: Parameter2
  type: string
  default: SomeParameterValue2

variables:
- name: Variable1
  value: SomeVariableValue1

- name: Variable2
  value: SomeVariableValue2

- ${{ if eq(parameters.Parameter1, 'SomeParameterValue1') }}:
  - name: VariableBasedUponParameter
    value: Parameter1 Equals SomeParameterValue1

- ${{ else }}:
  - name: VariableBasedUponParameter
    value: Parameter1 Does Not Equal SomeParameterValue1

jobs:
- job: Blah
  displayName: Blah
  steps:
  - ${{ if eq(parameters.Parameter1, 'SomeParameterValue1') }}:
    - script: |-
        echo Hello

  - ${{ if eq(variables['VariableBasedUponParameter'], 'Parameter1 Equals SomeParameterValue1') }}:
    - script: |-
        echo Hello again

  - script: |-
      echo $(VariableBasedUponParameter)

  - script: |-
      echo ${{ parameters.Parameter1 }}
```

but got:

```yaml
parameters:
- name: Parameter1
  displayName: SomeParameterValue1
  type: string

- name: Parameter2
  displayName: SomeParameterValue2
  type: string

variables:
- name: Variable1
  value: SomeVariableValue1

- name: Variable2
  value: SomeVariableValue2

- ${{ if eq('${{ parameters.Parameter1 }}', 'SomeParameterValue1') }}:
  - name: VariableBasedUponParameter
    value: Parameter1 Equals SomeParameterValue1

- ${{ else }}:
  - name: VariableBasedUponParameter
    value: Parameter1 Does Not Equal SomeParameterValue1

jobs:
- job: Blah
  displayName: Blah
  steps:
  - ${{ if eq('${{ parameters.Parameter1 }}', 'SomeParameterValue1') }}:
    - script: |-
        echo Hello

  - ${{ if eq('$(VariableBasedUponParameter)', 'Parameter1 Equals SomeParameterValue1') }}:
    - script: |-
        echo Hello again

  - script: |-
      echo $(VariableBasedUponParameter)

  - script: |-
      echo ${{ parameters.Parameter1 }}
```

Meir017:
expecting:
```yaml
- ${{ if eq(parameters.Parameter1, 'SomeParameterValue1') }}:
```

but getting:
```yaml
- ${{ if eq('${{ parameters.Parameter1 }}', 'SomeParameterValue1') }}:
```

expecting:
```yaml
- ${{ if eq(variables['VariableBasedUponParameter'], 'Parameter1 Equals SomeParameterValue1') }}:
```

but getting:
```yaml
- ${{ if eq('$(VariableBasedUponParameter)', 'Parameter1 Equals SomeParameterValue1') }}:
```

@premun is what we're getting here correct?

premun:
I believe the expectations are correct, not what we're getting

docs/AzureDevOps/GettingStarted.md
```diff
@@ -79,17 +79,19 @@ An example of a pipeline that builds and tests your PR can look like this:
 ```csharp
 ...
 
+private static readonly Variable s_DotnetVersionVariable = new Variable("DotnetVersion", string.Empty);
```
premun, line 82: I like this example but maybe we could just call it simply `DotnetVersion` to keep it nicer?

Meir017: done

src/Sharpliner/AzureDevOps/ConditionedExpressions/ConditionExtensions.cs
```diff
@@ -36,6 +36,14 @@ public static Conditioned<VariableBase> Variable(this IfCondition condition, str
     public static Conditioned<VariableBase> Variable(this IfCondition condition, string name, int value)
         => Conditioned.Link<VariableBase>(condition, new Variable(name, value));
 
+    /// <summary>
+    /// Defines a variable.
+    /// </summary>
+    /// <param name="condition">Conditioned definition</param>
+    /// <param name="variable">Variable instance</param>
+    public static Conditioned<VariableBase> Variable(this IfCondition condition, Variable variable)
```
premun, line 44: One more question/idea - does it make sense to change this to maybe accept the value as a third parameter too? So that we can just do `Variable(DotnetVersion, "6.0.100")`?

Meir017: I think we could have both, as now we could use `Variable(DotnetVersion with { Value = "6.0.100" })`

Meir017: I just tried this, this doesn't work since now the `Variable` has an implicit conversion to `string` and the overload of `Variable(this IfCondition condition, string name, string value)` is used

PR #374, opened 2024-11-18
feature: Support definition classes without parametersless constructor

### LINKED ISSUE #373: Support definition classes without parametersless constructor
The following code breaks the build:

```csharp
public class DemoPipeline(string component) : PipelineDefinition // <-- this breaks the build
{
    public override string TargetFile => "component/.pipelines/test.yml";
    public override Pipeline Pipeline => new Pipeline();
}

public abstract class ComponentPipelines(string component) : ISharplinerDefinitionCollection
{
    public IEnumerable<ISharplinerDefinition> Definitions =>
    [
        new DemoPipeline(component)
    ];
}

public class DemoServicePipelines() : ComponentPipelines("demo-service");
```

We want to have a group of pipeline definitions and then for each component we have generate multiple pipelines.

I current workarround is to make `DemoPipeline` abstract, and then extend from it (wanted to avoid this)

Meir017:
closes #373

### DIFF
--- a/src/Sharpliner/SharplinerPublisher.cs
+++ b/src/Sharpliner/SharplinerPublisher.cs
@@ -150,7 +150,7 @@ public class SharplinerPublisher(TaskLoggingHelper logger)
         var pipelines = new List<T>();
         var typeToFind = typeof(T);
 
-        foreach (Type type in assembly.GetTypes().Where(t => t.IsClass && !t.IsAbstract && t.IsAssignableTo(typeToFind)))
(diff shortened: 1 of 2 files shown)


premun (approved):
Ah, nice

premun:
Though technically you could have made the pipeline abstract to avoid this

Meir017:
> Though technically you could have made the pipeline abstract to avoid this

Due to C# not [anonymous classes](https://www.baeldung.com/java-anonymous-classes#2-implement-an-interface) (java example) I think a cleaner approach is often to create a template class with the required arguments passed into the ctor

PR #376, opened 2024-11-19
Feature: strongly typed templates

### LINKED ISSUE #356: Allow referncing templates in a strongly-typed way
Instead of using `TemplateParameters` it would be nice to have a strongly-typed parameters class.

currently:
```cs
.JobTemplate("template1.yml", new TemplateParameters
{
    { "enableTelemetry", true },
})
```

goal:
```cs
.JobTemplate<Template1>(new()
{
    EnableTelemetry = true
})
```

This achieves two things:
- Removing the explicit need to specify the template path
- Removing the explicit need for parameter names and dynamic types

Meir017:
closes #356


premun:
@Meir017 I thought we could have the strong-typed parameter types without source generators?

Let me try something

src/Sharpliner/AzureDevOps/TemplateDefinition.cs
```diff
@@ -257,3 +260,79 @@ internal TemplateDefinition()
     {
     }
 }
+
+public abstract class TemplateDefinition<T, TParameters> : TemplateDefinition<T> where TParameters : class, new()
+{
+    protected TParameters TypedParameters { get; private set; }
```
premun, line 266: Do we need this?

src/Sharpliner/AzureDevOps/TemplateDefinition.cs
```diff
@@ -257,3 +260,79 @@ internal TemplateDefinition()
...
+    }
+
+    protected TemplateDefinition(TParameters typedParameters)
+    {
+        TypedParameters = typedParameters;
+    }
+
+    public sealed override List<Parameter> Parameters => ToParameters(TypedParameters);
+
+    public TemplateParameters TemplateParameters => ToTemplateParameters(TypedParameters);
```
premun, line 280: Do we need this?

Meir017: we can probably inline this call in `public static implicit operator Template<T>(TemplateDefinition<T, TParameters> definition)`

src/Sharpliner/AzureDevOps/TemplateDefinition.cs
Meir017: We need this to be able to do 2 things
- `List<Parameter> Parameters => ToParameters(TypedParameters)`
- `public static implicit operator Template<T>(TemplateDefinition<T, TParameters> definition)`

Meir017: cleaned up

src/Sharpliner/AzureDevOps/TemplateDefinition.cs
Meir017: cleaned up

Meir017:
just needs a bit more docs

Meir017:
@premun ready now

premun:
@Meir017 what happens if we have a step parameter and give it a default valuet? Does this even work in YAML only?

Meir017:
@premun this is valid, see in https://github.com/sharpliner/sharpliner/blob/2b5fc6588df1cb6dfa4d329c02ed49cf2b60bdbf/tests/Sharpliner.Tests/AzureDevOps/TemplateTests.cs#L175 and a few more examples in the tests for parameters with default values.

premun:
Cool! I will review tomorrow, didn't get to it yet, but I think we'll merge almost as-is. I love the detailed xmldocs.

We should also redo the template guidance in the reference/readme.

I think this change is super valuable already without the generators

tests/Sharpliner.Tests/AzureDevOps/TemplateTests.cs
```diff
@@ -248,6 +447,237 @@ public void Stage_Template_Definition_Serialization_Test()
             """);
     }
 
+    private class Stage_Typed_Template_Definition(StageTypedParameters? parameters = null) : StageTemplateDefinition<StageTypedParameters>(parameters)
+    {
+      public override string TargetFile => "stage-template.yml";
+
+        public override ConditionedList<Stage> Definition => 
+        [
+          ..new Stage_Template_Definition().Definition,
```
premun, line 456: Quick question about this one: should we add an implicit cast so that just `new()` is enough?

Meir017: This is just me being lazy on the tests and reusing the definition

PR #379, opened 2024-11-25
feature: extends template definition

### LINKED ISSUE #377: feature request: Defining a template that extends another template
azure pipelines supports defining a template that extends another template.

for example:

`template-a.yml`:
```yml
parameters:
- name: paramA
  type: string
- name: paramB
  type: number

extends:
  template: `template-b.yml`:
  parameters:
...
```

`template-b.yml`:
```yml
parameters:
...
stages:
...
```

My issue here is that `public abstract class TemplateDefinition<T>` exposes a `public abstract ConditionedList<T> Definition { get; }` which will not work to render an `extends` object.

my current workarround is to define a type that extends the `ExtendsPipelineDefinition` type and set just the `Parameters` & `Extends` properties of the `PipelineWithExtends Pipeline` property

Meir017:
closes #377

### DIFF
--- a/src/Sharpliner/AzureDevOps/TemplateDefinition.cs
+++ b/src/Sharpliner/AzureDevOps/TemplateDefinition.cs
@@ -195,8 +195,8 @@ public abstract class TemplateDefinition : AzureDevOpsDefinition
 /// <summary>
 /// This is the ancestor of all definitions that produce a Azure pipelines template.
 /// </summary>
-/// <typeparam name="T">Type of the part of the pipeline that this template is for (one of stages, steps, jobs or variables)</typeparam>
-public abstract class TemplateDefinition<T> : TemplateDefinition, ISharplinerDefinition
+/// <typeparam name="T">Type of the part of the pipeline that this template is for (one of extends, stages, steps, jobs or variables)</typeparam>
+public abstract class TemplateDefinitionBase<T> : TemplateDefinition, ISharplinerDefinition
 {
     /// <summary>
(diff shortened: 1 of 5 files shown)


src/Sharpliner/AzureDevOps/TemplateDefinition.cs
```diff
@@ -220,7 +220,7 @@ public abstract class TemplateDefinition<T> : TemplateDefinition, ISharplinerDef
     /// Returns the definition of the template.
     /// </summary>
     [DisallowNull]
-    public abstract ConditionedList<T> Definition { get; }
+    public abstract T Definition { get; }
```
premun, line 223: Why did this change?

Meir017: I added `TemplateDefinitionBase<T>` which is the base class for `TemplateDefinition<ConditionedList<T>>`

this is in order to support templates of `extends` which contain two properties, `parameters` & `extends` but the `extends` is not a list, just a single object

premun: I will need to have a closer look to understand if it should be `Conditioned<T>`

Meir017:
I might have missed this, should we just extend `ExtendsPipelineDefinition<TPipeline>`? my main concern is that templates that extends other template can only declare the `parameters` & `extends` properties

premun:
I will have a look

premun (approved):
Had a closer look and it looks good!

Issue #378, opened 2024-11-24
feature request: allow passing variable & parameter to template parameter


Meir017:
would be nice to be able to pass variable/parameter to template parameters:

```csharp
private class MyTestPipeline : PipelineDefinition
{
   private static readonly Variable myVar = new Variable("myVar", "myValue");
   private static readonly Parameter myParam = StringParameter("myParam");
   public override Pipeline Pipeline => new()
   {
      Variables = [myVar],
      Parameters = [myParam],
      Stages =
      [
        StageTemplate("the-stage", new()
         {
            ["myParam"] = myParam,
            ["myVar"] = myVar,
         })
      ]
   };
}
```

generates the following YAML:

```yaml
parameters:
- name: myParam
  type: string
variables:
- name: myVar
  value: myValue
stages:
- template: the-stage.yml
  parameters:
    myParam: ${{ parameters.myParam }}
    myVar: ${{ variables.myVar }}
```

expected YAML:
```yaml
parameters:
- name: myParam
  type: string
variables:
- name: myVar
  value: myValue
stages:
- template: the-stage
  parameters:
    myParam:
      name: myParam
      type: string
    myVar:
      name: myVar
      value: myValue
```

Meir017:
I wonder if a better solution here would be to support something like

```csharp
StageTemplate("the-stage", new()
{
  ["myParam"] = parameters[myParam], // <-- this
  ["myVar"] = variables[myVar], // <-- this (still pending on https://github.com/sharpliner/sharpliner/issues/375)
})
```

by adding:

```diff
public class VariablesReference
{
+  public VariableReference this[Variable variable] => new(variable.Name);
}

public abstract class TemplateDefinition : AzureDevOpsDefinition
{
  public sealed class TemplateParameterReference
  {
+    public ParameterReference this[Parameter parameter] => new(parameter.Name);
  }
}
```

PR #380, opened 2024-11-25
feature: allow passing variable & parameter to template parameter

Meir017:
related to #378

### DIFF
--- a/tests/Sharpliner.Tests/AzureDevOps/ParameterReferenceTests.cs
+++ b/tests/Sharpliner.Tests/AzureDevOps/ParameterReferenceTests.cs
@@ -10,18 +10,25 @@ public class ParameterReferenceTests
 {
     private class ParameterReferenceTest_Template : StageTemplateDefinition
     {
+        private static readonly Parameter p_pool = ObjectParameter("pool", new()
+        {
+            { "vmImage", "windows-latest" }
+        });
+
+        private static readonly Parameter p_stages = StageListParameter("stages");
+        private static readonly Parameter p_jobs = JobListParameter("jobs");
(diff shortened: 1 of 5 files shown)


tests/Sharpliner.Tests/AzureDevOps/ParameterReferenceTests.cs
```diff
@@ -32,16 +39,16 @@ private class ParameterReferenceTest_Template : StageTemplateDefinition
                 {
                     new Job("Job_1")
                     {
-                        Pool = parameters["pool"],
+                        Pool = parameters[p_pool],
```
premun, line 42: It's a step forward but I'd say you'd rather expect to pass `p_pool` only, right? As that's a parameter already.

Meir017: Definitely!

premun: So if we put this onto `Variable`:
```
public static implicit operator Conditioned<Pool>(Parameter parameter)
	=> new ParameterReference(parameter.Name); // Possibly something else
```

It just works out of the box. Technically we might need to use some other overload of `Conditioned<T>` but basically this is something that we should consider. I am not sure how to make it so that we don't need to list all possible target types but if you think about it, a variable should be possible to be inserted just at every place like this

premun: Are we able to go all the way and allow passing the parameters directly?

Meir017: Trying that here #392

PR #382, opened 2024-11-27
feature: support passing enum values to variables

Meir017:
follow up on #372

TODO:
- [ ] add tests


tests/Sharpliner.Tests/AzureDevOps/PipelineVariableTests.cs
```diff
@@ -6,6 +6,12 @@ namespace Sharpliner.Tests.AzureDevOps;
 
 public class PipelineVariableTests
 {
+    private enum Configuration
+    {
+        Debug,
```
Meir017, line 11: let's add here an `YamlMember` attribute just to be sure it's treated

PR #384, opened 2024-11-27
Add NuGet tasks

### LINKED ISSUE #383: feature request: add nuget tasks
- `NuGetCommand@2` - https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/nuget-command-v2
- `NuGetAuthenticate@1` - https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/nuget-authenticate-v1

api changes:
```diff
public abstract class AzureDevOpsDefinition
{
+  protected static NuGetTaskBuilder NuGet { get; } = new();
}

+public class NuGetTaskBuilder
+{
+  public NuGetAuthenticateTask Authenticate(string[]? nuGetServiceConnections = null, bool forceReinstallCredentialProvider = false);
+  public NuGetRestoreBuilder Restore => new();
+  public NuGetPushBuilder Push => new();
+  public NuGetPackBuilder Pack => new();
+  public NuGetCustomBuilder Custom => new();
+}

+ public class NuGetAuthenticateTask : AzureDevOpsTask { ... }

// the rest of nuget task classes
```

Meir017:
Fixes #383

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/sharpliner/sharpliner/pull/384?shareId=9c5670f5-a6fb-4ad1-9870-c196dfe37d6c).


Meir017:
@premun all of the code here was generated using copilot-workspace!

premun:
Nice. What was the prompt? The JSON scheme and some instructions? Or you just started typing the class?

Meir017:
@premun I used https://copilot-workspace.githubnext.com/
accessed through the button on the issue
image

Meir017:
some of the properties are incorrect... let me fix this. but it was close, I wasn't explicit enough with the requirements for copilot...

src/Sharpliner/AzureDevOps/Model/Tasks/Builders/NuGetTaskBuilder.cs
```diff
@@ -0,0 +1,278 @@
...
+        /// </code>
+        /// </example>
+        /// </summary>
+        /// <returns>A <see cref="NuGetPushInternalCommandTask"/> instance.</returns>
+        public NuGetPushInternalCommandTask ToInternalFeed(string publishVstsFeed)
+        {
+            return new(publishVstsFeed)
+            {
+            };
+        }
```
premun, line 211:
```suggestion
        public NuGetPushInternalCommandTask ToInternalFeed(string publishVstsFeed) => new(publishVstsFeed);
```

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/NuGetAuthenticateTask.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="NuGetAuthenticateTask"/> class.
+        /// </summary>
+        public NuGetAuthenticateTask() : base("NuGetAuthenticate@1")
+        {
+        }
+
+        /// <summary>
+        /// Gets or sets the NuGet service connections.
```
premun, line 19:
```suggestion
        /// Gets or sets the NuGet service connection credentials for feeds outside this organization.
```

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/NuGetAuthenticateTask.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// </summary>
+        [YamlIgnore]
+        public string[]? NuGetServiceConnections
+        {
+            get => GetString("nuGetServiceConnections")?.Split(',');
+            init => SetProperty("nuGetServiceConnections", string.Join(",", value!));
+        }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to force reinstall the credential provider.
```
premun, line 29:
```suggestion
        /// Gets or sets a value indicating whether to force reinstall of the credential provider even if already installed. Default: false.
```

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/NuGetCommandTask.cs
```diff
@@ -0,0 +1,26 @@
+using YamlDotNet.Serialization;
+
+namespace Sharpliner.AzureDevOps.Tasks
+{
+    /// <summary>
+    /// Represents the abstract NuGetCommand@2 task in Azure DevOps pipelines.
```
premun, line 6:
```suggestion
    /// Represents the abstract NuGetCommand@2 task in Azure DevOps pipelines.
    /// Use this task to restore, pack, or push NuGet packages, or run a NuGet command.
    /// This task supports NuGet.org and authenticated feeds like Azure Artifacts and MyGet.
    /// This task also uses NuGet.exe and works with .NET Framework apps.
    /// For .NET Core and .NET Standard apps, use the .NET Core task.
```

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/NuGetPackCommandTask.cs
```diff
@@ -0,0 +1,298 @@
...
+
+        /// <summary>
+        /// The <c>X</c> in version <see href="http://semver.org/spec/v1.0.0.html">X.Y.Z</see>.
+        /// </summary>
+        [YamlIgnore]
+        public string MajorVersion
+        {
+            get => GetString("majorVersion")!;
+            init => SetProperty("majorVersion", value);
+        }
```
premun, line 209: I think all of these should rather be

```suggestion
        public string? MajorVersion
        {
            get => GetString("majorVersion");
            init => SetProperty("majorVersion", value);
        }
```

premun (commented):
Thanks again!

Looking at https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/nuget-command-v2?view=azure-pipelines, which is pretty horrible, this will be amazing to use.

Left some minor comments.

premun:
@Meir017 I'd add `Nuget.Authenticate` into the main example in the top level readme:

image

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/NuGetPackCommandTask.cs
Meir017: these version arguments are required will be set through the ctor, are you sure here? I think the value here is never null

premun: Ah, I see. For those coming from a ctor we shouldn't then, yeah, or we should maybe even do a nullcheck in the setter?

Meir017: do we have a standard way of requiring task properties to be non-null?

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/NuGetPackCommandTask.cs
```diff
@@ -0,0 +1,298 @@
...
+
+        /// <summary>
+        /// Gets or sets the pattern to search for csproj or nuspec files to pack.
+        /// </summary>
+        [YamlIgnore]
+        public string PackagesToPack
+        {
+            get => GetString("packagesToPack")!;
+            init => SetProperty("packagesToPack", value);
+        }
```
premun, line 47: This one is not required in the ctor so should be like this?

```suggestion
        public string? PackagesToPack
        {
            get => GetString("packagesToPack");
            init => SetProperty("packagesToPack", value);
        }
```

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/NuGetRestoreCommandTask.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// Specifies the credentials to use for external registries located in the selected <c>NuGet.config</c>.
+        /// This is the name of your NuGet service connection. 
+        /// For feeds in this organization or collection, leave this blank; the build's credentials are used automatically.
+        /// </summary>
+        [YamlIgnore]
+        public string ExternalFeedCredentials
+        {
+            get => GetString("externalFeedCredentials")!;
+            init => SetProperty("externalFeedCredentials", value);
+        }
```
premun, line 111:
```suggestion
        public string? ExternalFeedCredentials
        {
            get => GetString("externalFeedCredentials");
            init => SetProperty("externalFeedCredentials", value);
        } 
```

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/NuGetPackCommandTask.cs
premun: I don't think we do but if we have a setter on a required property, it should maybe throw? Because the YAML it will produce will technically be wrong?

Meir017: I think we could throw a ArgumentException

premun (approved):
Great, thanks!

PR #385, opened 2024-11-29
chore: remove duplicate readme

Meir017:


premun (approved):
I think the readme was different a bit because some things didn't make sense to be in a NuGet but I guess we don't want to maintain two copies anyway

PR #387, opened 2024-11-30
generate readme examples

Meir017:
@premun I think we could with a minimal docfx-like markdown rendered (to markdown, not html)

This will ensure our readme us accurate


eng/DocsGenerator/Program.cs
```diff
@@ -0,0 +1,89 @@
...
+                var language = line["[!code-".Length..line.IndexOf("[]")];
+
+                var codeSnippet = GetCodeSnippet(filepath, startLine, endLine, language);
+
+                builder.Append(codeSnippet);
+            }
+            else
+            {
+                builder.AppendLine(line);
+            }
```
premun, line 44: Personally a fan of early return:
```suggestion
            if (!line.StartsWith("[!code-"))
            {
                builder.AppendLine(line);
                continue;
            }

            var start = line.IndexOf('(');
            var end = line.IndexOf('#');
            var filepath = line.Substring(start + 1, end - start - 1);
            var lines = line.Substring(end + 1, line.Length - end - 1 - ")]".Length).Split('-');

            if (lines.Length != 2)
            {
                throw new ArgumentOutOfRangeException($"Invalid line range in {line}");
            }

            var startLine = int.Parse(lines[0].Substring("L".Length));
            var endLine = int.Parse(lines[1].Substring("L".Length));
            var language = line["[!code-".Length..line.IndexOf("[]")];

            var codeSnippet = GetCodeSnippet(filepath, startLine, endLine, language);

            builder.Append(codeSnippet);
```

premun (commented):
Interesting, more lightweight than the other proposal in your other repo. I like it and I guess it will work better but I think we also need a pipeline step verifying we have exported the documentation.

I guess we will also need some unindentation? Or we would just unindent the code in the source file?

Meir017:
> Interesting, more lightweight than the other proposal in your other repo. 
This solves part of the challenge, this only addresses the readme file, not the C# documentation comments

> I like it and I guess it will work better but I think we also need a pipeline step verifying we have exported the documentation.
> 
> I guess we will also need some unindentation? Or we would just unindent the code in the source file?
I'm not sure we need, the snippet-generator unindents all of the lines based on the first line assuming the first line will have the least amount of leading spaces

premun:
> I think we also need a pipeline step verifying we have exported the documentation.

I think we should try to call it in a PR build and fail if there's a change in `README.md`?

Meir017:
example failure -

image

eng/Sharpliner.CI/PullRequestPipeline.cs
```diff
@@ -27,6 +27,13 @@ class PullRequestPipeline : SingleStagePipelineDefinition
                 {
                     StepLibrary(new ProjectBuildSteps("src/**/*.csproj")),
 
+                    new DotNetCoreCliTask("run")
```
Meir017, line 30: will be cleaned up after #391

tests/Sharpliner.Tests/AzureDevOps/DocsTests.cs
```diff
@@ -0,0 +1,199 @@
...
+            },
+
+            DotNet.Build("src/MyProject.csproj") with
+            {
+                Timeout = TimeSpan.FromMinutes(20)
+            },
+        ];
+
+        var yaml = SharplinerSerializer.Serialize(steps);
+        // TODO: assert yaml
```
premun, line 95: Can we add this?

eng/DocsGenerator/README.template.md
```diff
@@ -0,0 +1,186 @@
...
+Be it a `${{ compile time parameter }}`, a `variable['used in runtime']` or a `$(macro)` syntax, with Sharpliner you won't have to worry about which one to pick as it understands the context and selects the right one for you.
+
+### Pipeline validation
+Your pipeline definition can be validated during publishing and you can uncover issues, such as typos inside `dependsOn`, you would only find by trying to run the pipeline in CI.
+This gives you a faster dev loop and greater productivity.
+
+We are continuously adding new validations as we find new error-prone spots.
+Each validation can be individually configured/silenced in case you don't wish to take advantage of these:
+
+[!code-csharp[](tests/Sharpliner.Tests/AzureDevOps/DocsTests.cs#L181-L198)]
```
premun, line 134: Looks really good but maybe we should just give it a named tag and look for that? I am afraid that it will be quite painful to keep moving these all the time in cases like we will need to add a `using` into the target file and now all blocks are one line lower.

So maybe we do  
`[!code-csharp[](tests/Sharpliner.Tests/AzureDevOps/DocsTests.cs#ValidationConfiguration)]`  
and then we will have

```csharp
// #ValidationConfiguration
...
// #ValidationConfiguration
```

Meir017: I wonder if it would be clearer to follow docfx's pattern of having the named tag point to a `#region <name>`

premun: These could be regions too, yeah

Meir017:
@premun The remaining documentation files will be added in a follow up PR

premun (approved):
Nice! I think this will be super useful

PR #394, opened 2024-12-07
docs: generate markdown docs

Meir017:


README.md
```diff
@@ -1,3 +1,10 @@
+<!--
```
premun, line 1: I don't think we should have these headers. The README might not be read raw, not in a rendered format, like in tools around NuGet packages, or just when looking in the repo.
If someone changes the README only and not the template, we will get a failed test so we're protected. Also there's not that many contributors apart from us and new ones will learn easily.

Meir017: done, removed this

eng/DocsGenerator/Program.cs
```diff
@@ -10,7 +11,24 @@ public class Program
...
     public static void Main(string[] args)
     {
-        var templatePath = GetRelativeToGitRoot("eng/DocsGenerator/README.template.md");
+        var failIfChanged = args.Contains("FailIfChanged=true");
+
+        var resources = new (string Template, string Output)[]
+        {
+            ("eng/DocsGenerator/README.template.md", "README.md"),
+            ("eng/DocsGenerator/GettingStarted.template.md", "docs/AzureDevOps/GettingStarted.md"),
+            ("eng/DocsGenerator/DefinitionReference.template.md", "docs/AzureDevOps/DefinitionReference.md"),
```
premun, line 20: Maybe later, but this would be cool if these were args too.
Maybe like this `templatePath:destinatioPath`:
```
eng/DocsGenerator/README.template.md:README.md
```

And for local runs, we'd have a script which would just `dotnet run` this project with the same args

premun (approved):

eng/DocsGenerator/Program.cs
Meir017: @premun when would you want to run just one of the files? This runs really fast

PR #397, opened 2024-12-13
test: use Verify for yaml validation

Meir017:
this could be a more simple way to write tests validation azure-pipeline YAML's

- [ ] remove PublicApiExporter.csproj


premun (approved):
Thanks. This idea already crossed my mind several times before but I never got around to it

tests/Sharpliner.Tests/TestSetup.cs
```diff
@@ -0,0 +1,22 @@
+using System.Runtime.CompilerServices;
+
+namespace Sharpliner.Tests;
+
+public class TestSetup
+{
+    [ModuleInitializer]
```
Meir017, line 7: this failes with a compile error, created https://stackoverflow.com/questions/79283695/having-a-net9-0-test-project-depend-on-a-library-that-depends-on-polyfill-doesn FYI @simoncropp

Meir017:
we should remove `PublicApiExporter.csproj` as part of this

tests/Sharpliner.Tests/TestSetup.cs
SimonCropp: is it this https://github.com/SimonCropp/Polyfill?tab=readme-ov-file#targetframeworks

As Polyfill is implemented as a source only nuget, the implementation for each polyfill is compiled into the IL of the resulting assembly. As a side-effect that implementation will continue to be used even if that assembly is executed in a runtime that has a more efficient implementation available.

As a result, in the context of a project producing nuget package, that project should target all frameworks from the lowest TargetFramework up to and including the current framework. This way the most performant implementation will be used for each runtime. Take the following examples:

If a nuget's minimum target is net6, then the resulting TargetFrameworks should also include net7.0 and net8.0
If a nuget's minimum target is net471, then the resulting TargetFrameworks should also include net472 and net48"

Meir017: Got it, so we would need to add additional frameworks here instead of targeting netstandard2.0.

Do we need to target netstandard 2.0 ? Or just the other frameworks?

Meir017: The test project is targeting net9.0, not sure I understand

premun (approved):

PR #398, opened 2024-12-13
feature: support netstandard2.1

Meir017:
not sure if we should make an effort for netstandard2.0 or netstandard2.1 which doesn't support netframework


Meir017:
closing in favor of #399

PR #399, opened 2024-12-14
feature: support netstandard2.0

Meir017:
see https://dotnet.microsoft.com/en-us/platform/dotnet-standard#versions

netstandard 2.0 will work for netframework as well
image


premun:
Thanks for putting in the work. It's interesting to see that it's not so bad to go down to `netstandard`. I think there is some benefit to this. I wonder how much baggage we will take on if we decide to keep supporting 2.0...

I guess we can try and see?

premun (approved):

PR #408, opened 2024-12-16
bugfix: name validation supporting parameter

### LINKED ISSUE #402: enhancement: Name Validation preventing using a parameter for stage name
consider the following code

```cs
public class DemoStageTemplate : StageTemplateDefinition
{
    public override TargetPathType TargetPathType => TargetPathType.RelativeToGitRoot;
    public override string TargetFile => "demo-template.yaml";

    public override List<Parameter> Parameters => 
    [
        StringParameter("stageName", defaultValue: "build_stage"),    
    ];

    public override ConditionedList<Stage> Definition => 
    [
        Stage(parameters["stageName"]) with
        {

        }
    ];
}
```

when building this produces the following error:

Sharpliner.targets(20,5): error :   Validation of definition DemoStageTemplate failed!
Sharpliner.targets(20,5): error :     - Invalid character found in name `${{ parameters.stageName }}`, only A-Z, a-z, 0-9, and underscore are allowed

after disabling the name-validating using a custom configuration, the generated yaml is correct:

```yaml
###
### DO NOT MODIFY THIS FILE!
###
### This YAML was auto-generated from DemoStageTemplate
### To make changes, change the C# definition and rebuild its project
###

parameters:
- name: stageName
  type: string
  default: build_stage

stages:
- stage: ${{ parameters.stageName }}

```

## Suggestion

Modify the `NameValidation` class to check if a name matches the parameters pattern `${{ parameters.<name> }}` and if so then validate the <name> parameter exists and is a string

Meir017:
closes #402


premun:
I think that maybe we should just disable this validation

Meir017:
Maybe we should just disable the name validation by default?

premun:
Yeah, let's do that

Meir017:
@premun I'm wondering if allowing compile/runtime expressions would fix issues here while still protecting us when having constant values for the name

premun:
You mean to remove everything between the start/end of these expressions and validating the rest?

Meir017:
@premun that's what I did here, basically if the name looks like an expression then ignore the validation

premun (commented):
Ah, I didn't see the update. 

Two notes: 
- there's also '$[ .. ]' for runtime, not sure it's applicable here
- could we instead reference the beginnings and end magic strings from the already existing constants somewhere in the codebase? It might be useful to have it referenced

premun (approved):

PR #410, opened 2024-12-18
feature: allow setting dependsOn to a parameters and support serializing as a single value when possible

### LINKED ISSUE #403: feature request: Setting stage.dependsOn to a parameters as a single value
the following doesn't compile

```cs
public class DemoStageTemplate : StageTemplateDefinition
{
    public override TargetPathType TargetPathType => TargetPathType.RelativeToGitRoot;
    public override string TargetFile => "demo-template.yaml";

    public override List<Parameter> Parameters => 
    [
        ObjectParameter<string>("dependsOn", defaultValue: [])
    ];

    public override ConditionedList<Stage> Definition => 
    [
        Stage("main-stage") with
        {
            DependsOn = parameters["dependsOn"], // <- this is invalid, changing to - [parameters["dependsOn"]] will generate an array of values
        }
    ];
}
```

goal for template yaml:

```yml
parameters:
- name: dependsOn
  type: object
  default: []

stages:
- stage: main-stage
  dependsOn: ${{ parameters.dependsOn }}
```

## Suggestion

Make the DependsOn be a new type that extends from `ConditionedList<string>` and has an implicit cast from `ParametersReference` & `string`.
Then, add a custom yaml-converter for this type then when the value is a list with a single item it serializes to a string instead of a array-of-string

Meir017:
closes #403

### DIFF
--- a/src/Sharpliner/AzureDevOps/AzureDevOpsDefinition.cs
+++ b/src/Sharpliner/AzureDevOps/AzureDevOpsDefinition.cs
@@ -882,7 +882,7 @@ public abstract class AzureDevOpsDefinition
     /// AzDO allows an empty dependsOn which then forces the stage/job to kick off in parallel.
     /// If dependsOn is omitted, stages/jobs run in the order they are defined.
     /// </summary>
-    protected static ConditionedList<string> NoDependsOn => new EmptyDependsOn();
+    protected static DependsOn NoDependsOn => [];
 
     #endregion
 }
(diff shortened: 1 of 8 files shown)


premun (approved):

premun (commented):
Does the `NoDependsOn` still work?

Meir017:
@premun I added a special serialization logic fir empty depends on instead of having a separate type

PR #411, opened 2024-12-19
fix NoDependsOn

Meir017:
follow up for #410

### DIFF
--- a/src/Sharpliner/AzureDevOps/AzureDevOpsDefinition.cs
+++ b/src/Sharpliner/AzureDevOps/AzureDevOpsDefinition.cs
@@ -882,7 +882,7 @@ public abstract class AzureDevOpsDefinition
     /// AzDO allows an empty dependsOn which then forces the stage/job to kick off in parallel.
     /// If dependsOn is omitted, stages/jobs run in the order they are defined.
     /// </summary>
-    protected static DependsOn NoDependsOn => [ string.Empty ];
+    protected static DependsOn NoDependsOn => [];
 
     #endregion
 }
--- a/src/Sharpliner/AzureDevOps/AzureDevOpsDefinition.cs
+++ b/src/Sharpliner/AzureDevOps/AzureDevOpsDefinition.cs
@@ -882,7 +882,7 @@ public abstract class AzureDevOpsDefinition
     /// AzDO allows an empty dependsOn which then forces the stage/job to kick off in parallel.
     /// If dependsOn is omitted, stages/jobs run in the order they are defined.
     /// </summary>
-    protected static DependsOn NoDependsOn => [];
+    protected static DependsOn NoDependsOn => [ string.Empty ];
 
     #endregion
 }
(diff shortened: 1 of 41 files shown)


src/Sharpliner/AzureDevOps/AzureDevOpsDefinition.cs
```diff
@@ -882,7 +882,7 @@ protected static Parameter StageListParameter(string name, string? displayName =
     /// AzDO allows an empty dependsOn which then forces the stage/job to kick off in parallel.
     /// If dependsOn is omitted, stages/jobs run in the order they are defined.
     /// </summary>
-    protected static DependsOn NoDependsOn => [];
+    protected static DependsOn NoDependsOn => [ string.Empty ];
```
premun, line 885: What didn't work there? It would be better, if supplying `[]` worked because I can imagine people might not even find the `NoDependsOn` and will just put `[]`

Meir017: I think we ignore empty collections

premun: Can't we tell the properties through the `YamlMember` attribute to not serialize `null` only and make these `null` by default? So that `[]`serializes as `[]`, Count == 1 serializes as string via the `IYamlConvertible` and no value makes the property not appear?

Meir017: https://github.com/sharpliner/sharpliner/blob/58c12072b8f24471ebfc66aa81e41b17c95359a2/src/Sharpliner/SharplinerSerializer.cs#L23

we globally ignore empty collections, I think that changing this to include empty collections will change too much

premun: I don't follow - my proposal is to go to stage and job where there are DependsOn members, making them nullable and changing their attribute by setting OmitWhenNull.
Won't that work?

Meir017: I got this to work, 

@premun I just looked at the docs, I wasn't able to find reference to having a `dependsOn: ''` I only found references showing having `dependsOn: []` to allow parallel

Meir017:
@premun I think this would be okay, since they will get a warning on this - `warning CS8670: Object or collection initializer implicitly dereferences possibly null member 'JobBase.DependsOn'`

tests/Sharpliner.Tests/Verified/AzureDevOps.Validation/DependsOnValidationTests.ConditionedDependsOn_Validation_Test.verified.txt
```diff
@@ -39,3 +39,9 @@ stages:
 
   - job: job_4
     dependsOn: ${{ parameters.jobDependsOn }}
+
+  - job: job_5
+    dependsOn: ''
```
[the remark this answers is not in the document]
Meir017: I wasn't sure about this, are `dependsOn: ''` & `dependsOn: []` equivalent? using the `[]` would be clearer (and cleaner code :) )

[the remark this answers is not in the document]
premun: I agree that '[]' is better. Can we do that then?

premun (approved):
Cool, thanks! I think it now behaves intuitively. I can publish a new version of the library. Or maybe after the name validation fix too?

Meir017:
@premun this was already merged on #408

Meir017:
@premun can you publish a new version? (maybe `1.7.0`)

premun:
Oops, I forgot to do it. 1.7.0 rolling out now

PR #414, opened 2024-12-19
tests: Nested conditioned blocks

### LINKED ISSUE #287: Nested conditioned blocks (`if`, `each`...) not working
It is impossible to nest plain conditioned blocks as only one condition is kept.

Meir017:
closes #287

### DIFF
--- a/tests/Sharpliner.Tests/AzureDevOps/ConditionedExpressions/ConditionalsTests.cs
+++ b/tests/Sharpliner.Tests/AzureDevOps/ConditionedExpressions/ConditionalsTests.cs
@@ -342,4 +342,36 @@ public class ConditionalsTests
 
         return Verify(pipeline.Serialize());
     }
+
+    class NestedConditionedBlocks : SingleStagePipelineDefinition
+    {
+        public override string TargetFile => "nested-conditioned-blocks.yml";
+        public override SingleStagePipeline Pipeline => new()
+        {
+            Jobs =
+            {
+                new Job("job1")
+                {
+                    Steps =
+                    {
+                        If.IsBranch("main")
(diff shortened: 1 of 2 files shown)


tests/Sharpliner.Tests/AzureDevOps/ConditionedExpressions/ConditionalsTests.cs
```diff
@@ -342,4 +342,36 @@ public Task IfCondition_And_InlineCondition_Test()
...
+            {
+                new Job("job1")
+                {
+                    Steps =
+                    {
+                        If.IsBranch("main")
+                            .Step(Bash.Inline("echo 'Main branch'"))
+                            .If.IsPullRequest
+                                .StepTemplate("template1.yml")
+                            .EndIf
```
premun, line 361: I think the thing that does not work is this

```suggestion
                        If.IsBranch("main")
                            .If.IsPullRequest
                                .StepTemplate("template1.yml")
                            .EndIf
```

Meir017: I'll add another test case for this

Meir017: got it, this doesn't compile right now... since the `IfCondition` type doesn't have an `If` property

premun: Yeah, it might just need a new API but maybe it will be more complicated to make it serialize properly. But maybe it's just about creating the `Conditioned` the right way

PR #415, opened 2024-12-21
feature: skip msbuild yaml generation

### LINKED ISSUE #400: Allow skipping msbuild yaml write task
To allow a gradual adoption for sharpliner we want to describe an existing YAML file by extending the base class `ExtendsTemplateDefinition<TParameters>` which is defined in an external repository

for example - using `v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates` (https://github.com/dotnet/xharness/blob/main/azure-pipelines.yml#L24 and many more repos)

Meir017:
closes #400

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,15 @@ Exchange YAML indentation problems for the type-safe environment of C# and let t
 
 ## Getting started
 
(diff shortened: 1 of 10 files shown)


premun (approved):
Thanks for also adding a test

PR #425, opened 2025-01-10
Allow Stage.DisplayName to be conditional

### LINKED ISSUE #424: Stage DisplayName is not conditioned
I want to be able to set the displayName of a stage conditionally. eg. 

```yaml
      - stage: thing
        ${{ if eq(parameters.WhatIf, true) }}:
          displayName: "Release: (WhatIf)"
        ${{ else }}:
          displayName: "Release"
```

The current type of DisplayName is string, 

https://github.com/sharpliner/sharpliner/blob/c2c1ed298335fad767b0b330e5b5fac05a1d4ac1/src/Sharpliner/AzureDevOps/Model/Stage.cs#L23C20-L23C32

whereas I think it needs to be `Conditioned<string>?` (like `Step` and `Job`)

flcdrg:
And add DisplayAs method

Fixes #424

### DIFF
--- a/src/Sharpliner/AzureDevOps/Model/Stage.cs
+++ b/src/Sharpliner/AzureDevOps/Model/Stage.cs
@@ -20,7 +20,7 @@ public record Stage : IDependsOn
     /// </summary>
     [YamlMember(Order = 2)]
     [DisallowNull]
-    public string? DisplayName { get; init; }
+    public Conditioned<string>? DisplayName { get; init; }
 
     /// <summary>
     /// The lock behavior of the stage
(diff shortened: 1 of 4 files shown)


src/Sharpliner/AzureDevOps/Model/Stage.cs
```diff
@@ -76,4 +76,9 @@ public Stage(string name, string? displayName = null)
             DisplayName = displayName;
         }
     }
+
+    /// <summary>
+    /// Sets the displayName property.
+    /// </summary>
+    public Stage DisplayAs(string displayName) => this with { DisplayName = displayName };
```
Meir017, line 83: Can we add a test that uses this?

flcdrg: done

premun:
Merging as we won't be able to do the conditioned scalar.

Thanks for the contribution, @flcdrg!

PR #436, opened 2025-01-26
feat: Use ParameterReference with TimeoutInMinutes Property

### LINKED ISSUE #435: Unable to Use ParameterReference with TimeoutInMinutes Property
Hey! First, thank you for the amazing work on Sharpliner! 💙

I really like the idea of passing the timeout as a TimeSpan and having it automatically converted into minutes in the generated YAML output. However, this design makes it impossible to use a **ParameterReference** for **TimeoutInMinutes** because it is a read-only property.

Here's an example of the YAML output I want to generate:
```yaml
parameters:
- name: task_name
  type: string
  default: NPM_task
- name: timeout_in_minutes
  type: number
  default: 30

steps:
- task: Npm@1
  name: ${{ parameters.task_name }}
  timeoutInMinutes: ${{ parameters.timeout_in_minutes }}
```

To achieve this, I tried the following Sharpliner code:

```csharp
public class Npm : StepTemplateDefinition
{
    public override string TargetFile => "tasks/externals/npm.yml";

    public override List<Parameter> Parameters => [
        new StringParameter("task_name", "NPM_task"),
        new NumberParameter("timeout_in_minutes", 30)
    ];

    public override ConditionedList<Step> Definition => [
        new AzureDevOpsTask("Npm@1")
        {
            Name = parameters["task_name"],
            TimeoutInMinutes = parameters["timeout_in_minutes"] // Error here
        }
    ];
}
```

However, I encountered the following error:
```
CS0200: The property or indexer 'Step.TimeoutInMinutes' cannot be assigned to because it is read-only.
```
### Question/Request
Is there a specific reason why TimeoutInMinutes is read-only? Since it’s a Conditioned<int>?, shouldn’t it accept a ParameterReference for dynamic assignment? Or is there a better approach I might be missing?

Thank you for considering this! Let me know if more details are needed :)

Meir017:
closes #435

### DIFF
--- a/src/Sharpliner/AzureDevOps/Model/Step.cs
+++ b/src/Sharpliner/AzureDevOps/Model/Step.cs
@@ -57,7 +57,8 @@ public abstract record Step
     public Conditioned<TimeSpan>? Timeout
     {
         get;
-        init {
+        init
+        {
             field = value;
             if (value?.Definition is not null)
             {
--- a/src/Sharpliner/AzureDevOps/Model/Step.cs
+++ b/src/Sharpliner/AzureDevOps/Model/Step.cs
@@ -54,13 +54,19 @@ public abstract record Step
     /// </summary>
     [YamlIgnore]
     [DisallowNull]
-    // public Conditioned<TimeSpan>? Timeout { get; init; }
     public Conditioned<TimeSpan>? Timeout
     {
         get;
         init {
             field = value;
-            TimeoutInMinutes = value?.Definition != null ? (int)value.Definition.TotalMinutes : null;
+            if (value?.Definition is not null)
+            {
+                TimeoutInMinutes = (int)value.Definition.TotalMinutes;
+            }
+            else if (value?.Condition is not null)
+            {
+                TimeoutInMinutes = new Conditioned<int>(default, value.Condition);
+            }
         }
     }
 
@@ -69,7 +75,6 @@ public abstract record Step
     /// </summary>
     [YamlMember(Order = 210)]
     public Conditioned<int>? TimeoutInMinutes { get; init; }
-    // public Conditioned<int>? TimeoutInMinutes => Timeout?.Definition != null ? (int)Timeout.Definition.TotalMinutes : null;
 
     /// <summary>
     /// A list of additional items to map into the process's environment.
--- a/src/Sharpliner/AzureDevOps/Model/Step.cs
+++ b/src/Sharpliner/AzureDevOps/Model/Step.cs
@@ -54,13 +54,22 @@ public abstract record Step
     /// </summary>
     [YamlIgnore]
     [DisallowNull]
-    public Conditioned<TimeSpan>? Timeout { get; init; }
+    // public Conditioned<TimeSpan>? Timeout { get; init; }
+    public Conditioned<TimeSpan>? Timeout
+    {
+        get;
+        init {
+            field = value;
+            TimeoutInMinutes = value?.Definition != null ? (int)value.Definition.TotalMinutes : null;
+        }
+    }
 
     /// <summary>
     /// Timeout after which the step will be stopped.
     /// </summary>
     [YamlMember(Order = 210)]
-    public Conditioned<int>? TimeoutInMinutes => Timeout?.Definition != null ? (int)Timeout.Definition.TotalMinutes : null;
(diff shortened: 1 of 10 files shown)


src/Sharpliner/AzureDevOps/Model/Step.cs
```diff
@@ -54,13 +54,22 @@ public abstract record Step
     /// </summary>
     [YamlIgnore]
     [DisallowNull]
-    public Conditioned<TimeSpan>? Timeout { get; init; }
+    // public Conditioned<TimeSpan>? Timeout { get; init; }
+    public Conditioned<TimeSpan>? Timeout
```
premun, line 58: I wonder if it's doable to just keep the Timeout visible to years? 
Maybe make the other internal?

So that when we set the Conditioned here, we would recreate the same one but return the int inside? Or that's too complicated?

Meir017: I'm not sure it's possible to have complex condition here

image
```cs
new InlineBashTask("./.dotnet/dotnet test src/MySolution.sln")
{
    DisplayName = "Run unit tests",
    ContinueOnError = true,
    TimeoutInMinutes = If.Equal(variables["main_run"], "true")
        .Value(30)
        .Else
        .Value(parameters["timeout_in_minutes"])
}
```
this doesn't compile

premun: Ok, so maybe not to have a full support for an if but what I meant is to not surface two different fields, I think that might be confusing. But rather redirect the `Conditioned<TimeSpan>` so that it serializes correctly (maybe we can somehow re-create the `Conditioned<int>` out of it?

premun: @Meir017 this works for instance:
```csharp
    public Conditioned<int>? TimeoutInMinutes => Timeout switch {
        ConditionedParameterReference<TimeSpan> parameter => new ConditionedParameterReference<int>(parameter.Parameter),
        ConditionedVariableReference<TimeSpan> variable => new ConditionedVariableReference<int>(variable.Variable),
        _ => Timeout?.Definition != null ? (int)Timeout.Definition.TotalMinutes : null
    };
```

But it's a quick solution, incomplete, sort of.

src/Sharpliner/AzureDevOps/Model/Step.cs
```diff
@@ -54,13 +54,28 @@ public abstract record Step
...
+        {
+            field = value;
+            if (value?.Definition is not null)
+            {
+                TimeoutInMinutes = (int)value.Definition.TotalMinutes;
+            }
+            else if (value?.Condition is not null)
+            {
+                TimeoutInMinutes = new Conditioned<int>(default, value.Condition);
+            }
```
Meir017, line 70: @premun this allows using simple expressions

src/Sharpliner/AzureDevOps/Model/Step.cs
```diff
@@ -54,13 +54,28 @@ public abstract record Step
...
+        }
+    }
 
     /// <summary>
     /// Timeout after which the step will be stopped.
     /// </summary>
     [YamlMember(Order = 210)]
-    public Conditioned<int>? TimeoutInMinutes => Timeout?.Definition != null ? (int)Timeout.Definition.TotalMinutes : null;
+    public Conditioned<int>? TimeoutInMinutes { get; init; }
```
premun, line 78: Should this be reverted then?

Meir017: I think this needs to stay as the `TimeoutInMinutes` property delegates to this property so here we save the actual value

premun: Can it be private init only?

Meir017: by having both properties have the `init` keyword we allow setting value to an integer for minutes or using a TimeSpan, the `get` here is used like any other property.

do you mean something like this?
```cs
public Conditioned<int>? TimeoutInMinutes { private get; init; }
```

premun: Ah, if we want both to be settable then why not, I just thought one would be settable for the user only

PR #443, opened 2025-04-10
feat: Added Job Dependency variables

jr-krandino:
I noticed that when doing a conditional that contained an [output variable from a job dependency](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#use-outputs-in-a-different-job), Sharpliner would attempt to wrap it in quotes.  For example, let's say I did the following statement:
`Equal("dependencies.A.outputs['ProduceVar.MyVar']", "true")`

This would result in:
`eq('dependencies.A.outputs['ProduceVar.MyVar']', true)`
As opposed to the intended:
`eq(dependencies.A.outputs['ProduceVar.MyVar'], true)`

To fix this, I've added a check to the "Conditions" object when checking to wrap quotes for tags that start with `dependencies.`, as well as a change to the SharplineSerializer which adds `dependencies` to the regex.

I've also added a `JobDependencyVariable` object which simplifies the process of creating these checks.  The above example can now be created with: `Equal(new JobDependencyVariable("MyVar", "ProduceVar", "A"), "true")`


Meir017:
can you add some unit-tests for this? I'm not sure what the usage would look like

jr-krandino:
Yeah, I can look into doing that.  I confess I'm not an expert when it comes to UnitTests, so it may take some time.  I also noticed that the APIExport test is failing.  Is there a process I'm supposed to run to update the verified file, or is it meant to be updated manually to match?

premun:
Hi, @jr-krandino, thanks for the contribution!

This is a great change you're proposing and it will make the library much better. But I agree tests would be nice.
Feel free to copy already existing tests (there should be some for variable usage which are probably similar).

For the API export test, there used to be a script to overwrite the file with the new API contents but I can't find it anymore.
I guess now you just need to copy the contents of the verified file into the expect one that gets opened if you run the test locally (details are in the exception). And then you commit this new file.

jr-krandino:
I've updated the API Export verification text, and added a test for `JobDependencyVariableReference` called `JobDependencyVariableReferenceTests`.

premun:
@jr-krandino would it be possible to make a nicer API to use this new dependency stuff? Similarly to how the `parameters` and `variables` are used? I think `new JobDependencyVariable(sharedVariableName, setterTaskName, setterJobName)` is not in line with how the rest of the APIs are designed.

Maybe it would be something like `dependencies["job"]["task"]["variable"]`?

jr-krandino:
Interesting point.  Something like this will take a massive rework.  I can't promise this is something I can achieve today, but I'll look into this when I have a little more free time.

jr-krandino:
I think I see a general way forward!  But in doing so, now I want to add some additional functionality.  Depending on how busy my work day is, I may be able to accomplish this today after all.  Will update if I end up blocked, thanks!

jr-krandino:
Not quite.  The issue isn't that there are cases where dependencies have different numbers of arguments, that's actually the ideal situation!  The issue is that there are some cases where dependencies have the **_same_** number of arguments, however, the expected input for them is different.

In the table posted above, you'll notice that both `stage to stage dependency (different stages)` and `Job to stage dependency (different stages)` take in the exact same number of arguments, however, their YAML output is completely different.  They are meant to be used in different areas of the YAML code, and under different situations.  Sadly, it makes for a very tricky solution, as the YAML output for them is _**VERY**_ similar, but just different enough that everything would break if used in the wrong spot.  So, because of those above use the `stageDependencies` object, and have the same number of variables, we unfortunately can't just give them both `stageDependencies["stage"]["job"]["step"]["variable"]`, as they have different outputs.

This is why I'm thinking we may have to disregard the standard yaml approach to this one.  We could make one of them `dependencies["stage"]["job"]["step"]["variable"]` instead, but then we get a conflict with `Stage to stage dependency (deployment job)`.  That's why I propose we do that AND create a 3rd `deploymentDependencies` object which would be for the two deployment job based ones.  It's kind of messy, and disregards the yaml approach, but it's the best way forward I can personally see.

Meir017:
based on my confusion of going through the `YAML Output` I think this is amazing!

src/Sharpliner/AzureDevOps/ConditionedExpressions/DependencyVariableReference.cs
```diff
@@ -0,0 +1,252 @@
...
+    /// <inheritdoc/>
+    public override string RuntimeExpression => $"dependencies.{JobName}.outputs['{StepName}.{VariableName}']";
+
+    /// <inheritdoc/>
+    public override string MacroExpression => $"$({StepName}.{VariableName})";
+
+    /// <inheritdoc/>
+    public override bool Equals(object? obj)
+    {
+        if (obj is DependencyVariableReference other)
```
Meir017, line 40: can be shortened to 
```cs
return (obj is DependencyVariableReference other) &&
    VariableName == other.VariableName &&
    JobName == other.JobName &&
    StepName == other.StepName);
```

tests/Sharpliner.Tests/AzureDevOps/DependencyVariableReferenceTests.cs
```diff
@@ -0,0 +1,390 @@
...
+                        {
+                            Script.Inline($"echo hello from Job {getterJobName}")
+                        }
+                    },
+                }
+            }
+        ];
+    }
+
+    // This test should create a variable in one job and reference it in the condition for another stage.
```
Meir017, line 343: consider removing the comments on these methods, they don't provide additional information to the test names :)

Meir017:
I think that it would be good to follow up and generate an integration test as an azure-pipelines that has multiple dummy stages and jobs to demonstrate the output referencing syntax

src/Sharpliner/AzureDevOps/ConditionedExpressions/DependencyVariableReference.cs
jr-krandino: Great point!  Done.

premun:
This is really cool! Thanks for putting in the work. 

I believe the proposed syntax is fine if we include the overview table from your last comment in the reference documentation. That table is really readable and people will hopefully be able to navigate through their problem using it. 

I will have another look later today. Thanks to @Meir017 for the reviews too!

jr-krandino:
Thanks, glad you all agree with the approach!  I can take a minute today to to write some documentation on this.  Is there any pre-existing documentation you'd like me to add this to, or is it dealer's choice?

Meir017:
for the docs, we have the source markdown files here https://github.com/sharpliner/sharpliner/tree/main/eng/DocsGenerator and link to real C# code from the unit-test project, see examples here https://github.com/sharpliner/sharpliner/tree/main/tests/Sharpliner.Tests/AzureDevOps/Docs

premun (dismissed):
I think the PR is high quality, the XML docs and the tests are also pretty thorough. Thanks for being this diligent!

I'd wait for the documentation update - fell free to just add a section in the [AzDO definition reference](https://github.com/sharpliner/sharpliner/blob/main/docs/AzureDevOps/DefinitionReference.md) - ideally using the doc generators that Meir linked.

Other than that I think we're good to merge!

jr-krandino:
Documentation has been added!

Meir017 (approved):

tests/Sharpliner.Tests/AzureDevOps/DependencyVariableReferenceTests.cs
```diff
@@ -0,0 +1,386 @@
...
+                    new Job(getterJobName)
+                    {
+                        Condition = Equal(dependencies.job.deploy[setterStageName, setterJobName, setterStepName, sharedVariableName], "true"),
+                        Steps =
+                        {
+                            Script.Inline($"echo hello from Job {getterJobName}")
+                        }
+                    },
+                }
+            }
```
premun, line 210:
```suggestion
            new Stage(getterStageName)
            {
                DependsOn = setterStageName,
                Jobs =
                {
                    new Job(getterJobName)
                    {
                        Condition = Equal(dependencies.job.deploy[setterStageName, setterJobName, setterStepName, sharedVariableName], "true"),
                        Steps =
                        {
                            Script.Inline($"echo hello from Job {getterJobName}")
                        }
                    },
                }
            }
```

docs/AzureDevOps/DefinitionReference.md
```diff
@@ -243,6 +244,40 @@ public override SingleStagePipeline Pipeline => new()
...
 
+### Dependency Variables
+
+If your variable is defined in another job, you can use the dependencies shorthand to refrence them.  Azure DevOps has [many different possible yaml outputs](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#dependency-syntax-overview) for referencing dependency variables, which is simplified by our shorthand syntax: `dependencies.<job|stage>.<deploy?>[]`.  Simply specify whether your reference is meant to live inside a `job` entry or a `stage` entry, then if the reference was instantiated within a deploy job, you simply add `.deploy` to the chain.
+
+For example, if you wanted to reference a variable created in a deployment job inside another stage's job entry, you'd do so like this.
+
+```csharp
+new Stage(getterStageName) {
+    DependsOn = setterStageName,
```
premun, line 255:
```suggestion
new Stage(getterStageName)
{
    DependsOn = setterStageName,
```

premun (approved):
I think the table in your documentation is even more understandable than the original AzDO documentation 😀

premun (approved):

premun:
🙏 Thank you for the contribution @jr-krandino, the changes are available in the version `1.7.7` of the Sharpliner package

PR #444, opened 2025-05-01
Fix: Environment tags in deploy jobs

jr-krandino:
Currently, when making a deploy job and setting the environment, if we set Tags, they are set like this:

```
environment:
  tags:
  - tag1
  - tag2
```

The issue is, Azure expects this to be a comma delimited string.  This fix makes it so we would instead get this:

```
environment:
  tags: tag1, tag2
```

### DIFF
--- a/src/Sharpliner/AzureDevOps/Model/Environment.cs
+++ b/src/Sharpliner/AzureDevOps/Model/Environment.cs
@@ -35,7 +35,14 @@ public record Environment
     /// <summary>
     /// Tag names to filter the resources in the environment
     /// </summary>
+    [YamlIgnore]
     public List<string> Tags { get; init; } = [];
+
+    /// <summary>
+    /// Internal property to serialize the tags as a comma-separated string.
+    /// </summary>
+    [YamlMember(Alias = "tags")]
+    public string? _Tags => Tags.Count > 0 ? string.Join(",", Tags) : null;
 
     /// <summary>
     /// Instantiates a new <see cref="Environment"/> with the specified name and optional resource name.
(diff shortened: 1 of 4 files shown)


src/Sharpliner/AzureDevOps/Model/Environment.cs
```diff
@@ -35,8 +35,15 @@ public record Environment
...
     /// Tag names to filter the resources in the environment
     /// </summary>
+    [YamlIgnore]
     public List<string> Tags { get; init; } = [];
 
+    /// <summary>
+    /// Internal property to serialize the tags as a comma-separated string.
+    /// </summary>
+    [YamlMember(Alias = "tags")]
+    public string? _Tags => Tags.Count > 0 ? string.Join(",", Tags) : null;
```
Meir017, line 45: this should be internal

jr-krandino: Interesting.  I modeled these after the approach taken in `DownloadTask.cs`.  That one was not internal.  Should it also be?

https://github.com/sharpliner/sharpliner/blob/9de870d282581478ccfce1c078f64789bd6c9470/src/Sharpliner/AzureDevOps/Model/Tasks/DownloadTask.cs#L48-L58

premun: I just wanted to post the same comment. But it might be worth setting the tags in some random unit test and test it too?

jr-krandino: I tested making it internal.  By doing that, it no longer outputs the yaml.  I think this is the same reason why it's public for the DownloadTask, if it isn't, the yaml gets ignored.

jr-krandino: I modified one of the template tests that has a deploy job in it to have some environmental tags as well.

premun: Ah, good to know. Probably hat was the reason! Thanks for trying though!

premun (approved):

PR #445, opened 2025-05-01
Fixed: Git Relative Templates Not Referenced Correctly

jr-krandino:
Currently, in Azure DevOps when we are referencing TemplateDefinitions, we are not taking into account if the file path of the template definition is meant to be relative to the git root.  This means when we reference a template that is being saved in `template/run-git.yaml`, and we attempt to reference it later in a pipeline that exists at `yaml/main-pipeline.yaml`, we create this output in `main-pipeline.yaml`:
```
    - template: template/run-git.yaml
```

The reason this is an issue is because when we run `yaml/main-pipeline.yaml`, it will attempt to call that template relative to itself, starting at `yaml`, leading to the wrong area.  To fix this, when a template is relative to the git root, Azure DevOps expects it to begin with `/`, looking like this:

```
    - template: /template/run-git.yaml
```

This fix makes it so, if a TemplateDefinition is set to be Relative to the Git Root, it will add a '/' to its filename when serialized.

### DIFF
--- a/src/Sharpliner/AzureDevOps/PublicDefinitions.cs
+++ b/src/Sharpliner/AzureDevOps/PublicDefinitions.cs
@@ -131,7 +131,14 @@ public abstract class ExtendsTemplateDefinition<TParameters> : ExtendsTemplateDe
     /// <param name="definition">The typed template definition</param>
     public static implicit operator Template<Extends>(ExtendsTemplateDefinition<TParameters> definition)
     {
-        return new Template<Extends>(definition.TargetFile, TypedTemplateUtils<TParameters>.ToTemplateParameters(definition._typedParameters));
+        string path = definition.TargetFile;
+
+        if (definition.TargetPathType == TargetPathType.RelativeToGitRoot)
+        {
+            path = "/" + path;
+        }
+
+        return new Template<Extends>(path, TypedTemplateUtils<TParameters>.ToTemplateParameters(definition._typedParameters));
     }
 }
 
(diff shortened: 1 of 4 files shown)


Meir017 (approved):

src/Sharpliner/AzureDevOps/PublicDefinitions.cs
```diff
@@ -131,7 +131,14 @@ protected ExtendsTemplateDefinition(TParameters? typedParameters = null)
     /// <param name="definition">The typed template definition</param>
     public static implicit operator Template<Extends>(ExtendsTemplateDefinition<TParameters> definition)
     {
-        return new Template<Extends>(definition.TargetFile, TypedTemplateUtils<TParameters>.ToTemplateParameters(definition._typedParameters));
+        string path = definition.TargetFile;
+
+        if (definition.TargetPathType == TargetPathType.RelativeToGitRoot)
+        {
+            path = "/" + path;
```
premun, line 138: Should we check if `path` already starts with `/`?

premun (commented):
Thank you for the contribution - just a small thing

src/Sharpliner/AzureDevOps/PublicDefinitions.cs
jr-krandino: So, funny enough, that was causing me a huge headache.  Originally I thought I could fix this by simply putting "/" at the start of the file name.  In actuality what that did was make the publisher create the `.yaml` files in the root of my C: Drive!  From seeing that, I'd actually think the better approach would be to prevent people from having a "/" at the start of their file name (perhaps by putting a `.TrimStart('/') in it when it's referenced or something?) but that was beyond the scope of what I was going for here.

I can add the check, but it'd end up being redundant if we go that route.  Thoughts?

premun: Ah, if that's the behaviour then let's keep it as is. Also possibly `//` does not matter hopefully

premun (approved):

premun:
@jr-krandino thank you for the contribution, there will be a new version `1.7.8` pushed shortly

PR #446, opened 2025-05-02
Elseif condition

jr-krandino:
While working with Sharpliner I noticed there was no way to invoke an ${{ elseif }} call.  This update adds .ElseIf as a Conditioned object, which will allow for elseifs like so:

```
If.Equal("A", "B")
  .Pool(new HostedPool("pool-A")
  {
      Demands = { "SomeProperty -equals SomeValue" }
  })
  .ElseIf.Equal("C", "D")
  .Pool(new HostedPool("pool-B"))
```

outputs:
```
${{ if eq('A', 'B') }}:
  name: pool-A
  demands:
  - SomeProperty -equals SomeValue
${{ elseif eq('C', 'D') }}:
  name: pool-B
```

I've updated a test that was meant to do ElseIf statements to utilize this.

### DIFF
--- a/src/Sharpliner/AzureDevOps/IfCondition.cs
+++ b/src/Sharpliner/AzureDevOps/IfCondition.cs
@@ -7,11 +7,17 @@ namespace Sharpliner.AzureDevOps;
 /// </summary>
 public abstract class IfCondition : Condition
 {
+    internal bool IsElseIf = false;
+
+    internal override string TagStart => IsElseIf ? ElseIfTagStart : IfTagStart;
+
     private static readonly (string Start, string End)[] s_tagsToRemove =
     [
         (IfTagStart, ExpressionEnd),
+        (ElseIfTagStart, ExpressionEnd),
         (ElseTagStart, ExpressionEnd),
         ('\'' + IfTagStart, ExpressionEnd + '\''),
+        ('\'' + ElseIfTagStart, ExpressionEnd + '\''),
         ('\'' + ElseTagStart, ExpressionEnd + '\''),
         (ExpressionStart, ExpressionEnd),
         ('\'' + ExpressionStart, ExpressionEnd + '\''),
@@ -41,8 +47,8 @@ public abstract class IfCondition : Condition
         return condition;
     }
 
-    internal static string WrapTag(string condition) =>
-        IfTagStart + condition + ExpressionEnd;
+    internal string WrapTag(string condition) =>
+        TagStart + condition + ExpressionEnd;
 }
 
 /// <summary>
--- a/src/Sharpliner/AzureDevOps/AzureDevOpsDefinition.cs
+++ b/src/Sharpliner/AzureDevOps/AzureDevOpsDefinition.cs
@@ -537,7 +537,7 @@ public abstract class AzureDevOpsDefinition
     ///   - group: prod
     /// </code>
     /// </summary>
-    protected static IfConditionBuilder If => new();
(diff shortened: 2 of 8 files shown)


src/Sharpliner/AzureDevOps/Model/Environment.cs
```diff
@@ -35,8 +35,15 @@ public record Environment
...
     /// Tag names to filter the resources in the environment
     /// </summary>
+    [YamlIgnore]
     public List<string> Tags { get; init; } = [];
 
+    /// <summary>
+    /// Internal property to serialize the tags as a comma-separated string.
+    /// </summary>
+    [YamlMember(Alias = "tags")]
+    public string? _Tags => Tags.Count > 0 ? string.Join(",", Tags) : null;
```
premun, line 45: Should this be part of the PR?

src/Sharpliner/AzureDevOps/TemplateDefinition.cs
```diff
@@ -195,7 +195,14 @@ internal TemplateDefinition(TParameters? typedParameters = null)
     /// <param name="definition">The typed template definition</param>
     public static implicit operator Template<T>(TemplateDefinition<T, TParameters> definition)
     {
-        return new Template<T>(definition.TargetFile, TypedTemplateUtils<TParameters>.ToTemplateParameters(definition._typedParameters));
+        string path = definition.TargetFile;
+
+        if (definition.TargetPathType == TargetPathType.RelativeToGitRoot)
+        {
+            path = "/" + path;
```
premun, line 202: I think this one also creeped in by mistake?

src/Sharpliner/AzureDevOps/IfCondition.cs
```diff
@@ -7,11 +7,17 @@ namespace Sharpliner.AzureDevOps;
 /// </summary>
 public abstract class IfCondition : Condition
 {
+    internal bool IsElseIf = false;
```
premun, line 10: Can you not make this a boolean flag but instead use inheritance and possibly override some of the fields such as `TagStart`?
Can you do the same for the builder?

premun (commented):
@jr-krandino I like the change, thank you for the contribution, just a couple of small comments

jr-krandino:
The excess code from another PR should all be removed now.

src/Sharpliner/AzureDevOps/IfCondition.cs
premun: @jr-krandino how about this one?

jr-krandino:
Hey @premun.  I think I'm having some sort of strange issue where my responses to your previous question about inheritance wasn't going out.  They're weirdly marked as "Pending".

441224090-68103d5e-7938-443c-af80-ff1e4b425c8d

Here's how it looks on my end, as well as the explanation on why I went the route I did.

premun:
@jr-krandino you probably made them in the "Files" tab and you "Started a review" so it's buffering all your comments until you publish the review.

premun (approved):

premun:
> Hey @premun. I think I'm having some sort of strange issue where my responses to your previous question about inheritance wasn't going out. They're weirdly marked as "Pending".
> 
> 441224090-68103d5e-7938-443c-af80-ff1e4b425c8d
> 
> Here's how it looks on my end, as well as the explanation on why I went the route I did.

I see, let's merge then. Thanks for this one!

PR #447, opened 2025-05-06
Support more built-in Azure DevOps tasks

mqmcgaha:
This PR adds support for the following currently unmodeled tasks:
- AzureCLI@2
- ExtractFiles@1
- UniversalPackages@0 (download)
- UniversalPackages@0 (publish)


Meir017:
I wonder about adding Pipeline task shorthands utility methods (like we have here https://github.com/mqmcgaha/sharpliner/blob/bf023982e1144b2577c4fdf03de59e62768c89ad/src/Sharpliner/AzureDevOps/AzureDevOpsDefinition.cs#L254-L309)

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/DownloadPipelineArtifactTask.cs
```diff
@@ -0,0 +1,208 @@
...
+    public string? TargetPath
+    {
+        get => GetString("targetPath");
+        init => SetProperty("targetPath", value);
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="DownloadPipelineArtifactTask"/> class with required properties.
+    /// </summary>
+    public DownloadPipelineArtifactTask()
```
Meir017, line 162: this task is already covered by https://github.com/sharpliner/sharpliner/blob/main/src/Sharpliner/AzureDevOps/Model/Tasks/DownloadTask.cs

Meir017 (changes requested):

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/DownloadPipelineArtifactTask.cs
mqmcgaha: Thanks - I see the implementation for current and specific download tasks

mqmcgaha:
> I wonder about adding Pipeline task shorthands utility methods (like we have here https://github.com/mqmcgaha/sharpliner/blob/bf023982e1144b2577c4fdf03de59e62768c89ad/src/Sharpliner/AzureDevOps/AzureDevOpsDefinition.cs#L254-L309)

Added a task builder for Azure CLI task. Made sense for this task

src/Sharpliner/AzureDevOps/Model/Tasks/Marketplace/AzureCliTask.cs
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    /// <summary>
+    /// Use when scriptType = ps || scriptType = pscore.
+    /// If this input is false, the line if ((Test-Path -LiteralPath variable:\LASTEXITCODE)) { exit $LASTEXITCODE } is appended to the end of your script.
+    /// This will propagate the last exit code from an external command as the exit code of PowerShell.
+    /// Otherwise, the line is not appended to the end of your script.
+    /// </summary>
+    [YamlIgnore]
+    public bool PowerShellIgnoreLASTEXITCODE
```
Meir017, line 113: this property and `PowerShellErrorActionPreference` are only relevant when "scriptType = ps || scriptType = pscore".

Consider having extended record classes to help with the validations here where these properties will exist on a more specific implementation that uses powershell

Meir017 (approved):
just some thoughts on the azure cli task where we should probably have more typed enforcements around which properties are visible based on the script type

premun:
> just some thoughts on the azure cli task where we should probably have more typed enforcements around which properties are visible based on the script type

This would be even more valuable - to have a branched builder like we have for the other tasks which then make it super easy and safe to use these tasks.

tests/Sharpliner.Tests/Verified/AzureDevOps/TaskBuilderTests.Serialize_AzureCli_Builder_Test.verified.txt
```diff
@@ -0,0 +1,44 @@
...
+    inputs:
+      azureSubscription: connectedServiceNameARM
+      scriptType: ps
+      scriptLocation: inlineScript
+      inlineScript: |+
+        Set-ErrorActionPreference Stop
+        Write-Host "Lorem ipsum dolor sit amet"
+
+  - task: AzureCLI@2
+    displayName: Write-Host "test"
```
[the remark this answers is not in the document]
mqmcgaha: Fixed

premun (approved):
Thank you! Will publish a new version of the package with these tasks shortly

premun (approved):

PR #453, opened 2025-05-15
docs: nicer marketplace tasks and add nuget link for jfrog

Meir017:
see here https://github.com/Meir017/sharpliner/blob/docs/definition-reference/docs/AzureDevOps/DefinitionReference.md#marketplace-tasks

image

FYI @flcdrg


premun (approved):
Nice, thanks!

Issue #450, opened 2025-05-08
Allow variable/parameter expressions for primitive AzureDevOpsTask inputs


trevorlacey-msft:
Similar to #267 , I'd like to be able to parameterize inputs for tasks that inherit from `AzureDevOpsTask`, specifically inputs that are `int` or `bool`.

For example, `CopyFilesTask` has this property:
```cs
[YamlIgnore]
public bool CleanTargetFolder
{
    get => GetBool("CleanTargetFolder", false);
    init => SetProperty("CleanTargetFolder", value);
}
```

As far as I can tell, there's no way for me to parameterize this (e.g. with `BooleanParameter`) because it only supports `bool`. In #267, the answer was to switch everything to `Conditioned<T>`, but that was only done for non `AzureDevOpsTasks` tasks.

premun:
Thanks for the suggestion, I agree that probably this work needs to be done for all properties across the repo.
I will see if Copilot can do that.

premun:
Okay, so this is possible and it works: https://github.com/premun/sharpliner/commit/8df2f5d64359b9bdc8eb7a26fd4ca2f41d5cfc2a

And Copilot seems to be doing a very good job at this

premun:
Okay, I hit an interesting issue:

- Some fields have default value, e.g. `enabled: true` on every task
- Default values are not serialized into YAML using the `[DefaultValue(...)]` attribute
- If we make `Conditioned<bool> Enabled` instead of `bool Enabled`, it will start to get serialized because it won't compare against `true` well
- We can override `Equals()` on `Conditioned<T>` because it is a `record` and it has the Equals implemented by C# (the member comparison one)

We can maybe now change all fields but the ones with default values first and figure out what to do about it later.

premun:
What we could do to fix the problem above is making these fields nullable:
- `Conditioned<bool>? Enabled { get; set; } = null;`
- Then it will not get serialized by default
- If someone assigns `true`, it will get serialized but maybe that's desired because it will match the C# better?
- It might be a breaking change as people's YAML will receive these previously hidden defaults (so should not break the YAML itself, just it will be dirty)

premun:
Okay, made some progress. I need to add a couple of tests but apart from that we're almost 100% covered:
https://github.com/sharpliner/sharpliner/pull/454

premun:
[...]

PR #454, opened 2025-05-24
Allow usage of expression in almost all fields

premun:
https://github.com/sharpliner/sharpliner/issues/450

### DIFF
--- a/src/Sharpliner/AzureDevOps/ConditionedExpressions/Conditioned.cs
+++ b/src/Sharpliner/AzureDevOps/ConditionedExpressions/Conditioned.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using YamlDotNet.Core;
 using YamlDotNet.Core.Events;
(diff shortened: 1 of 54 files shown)


docs/AzureDevOps/GettingStarted.md
```diff
@@ -176,14 +179,15 @@ jobs:
...
+        Write-Host 'Hello'
+        Write-Host 'World'
       displayName: Hello world
 
   - task: UseDotNet@2
     displayName: Install .NET SDK
     inputs:
       packageType: sdk
-      version: $(DotnetVersion)
+      version: ${{ variables['DotnetVersion'] }}
```
premun, line 190: I am not sure this is the correct syntax

premun:
@Meir017 this turned out to be a very large re-write. But I guess it's more or less backwards compatible.

However, I was thinking that `Conditioned` could be renamed to `Expression` since we're already doing a big change.

Thoughts?

Meir017:
hmm, `Expression` sounds good. I wonder if something like `AzDoExpression` or something with a bit context would be better, or maybe `AdoExpression`.

I'm leaning towards `AdoExpression`

PR #457, opened 2025-06-05
Add net10.0 TFM

premun:


.github/workflows/codeql-analysis.yml
Meir017, line 1: One day github actions will be supported too ;)

PR #458, opened 2025-06-05
Rename `Conditioned` to `AdoExpression`

premun:


premun:
@Meir017 this is a breaking change because for template/libraries you need to override memebers which used to be `ConditionedList`. E.g.

```csharp
class ProjectBuildSteps(string project) : StepLibrary
{
    public const string PackagePath = "artifacts/packages";

    public override List<AdoExpression<Step>> Steps =>...
}
```

I might add that type back just for backwards compatibility

README.md
```diff
@@ -198,7 +198,7 @@ Apart from obvious C# code reuse, you can also define sets of C# building blocks
 ```csharp
 class ProjectBuildSteps : StepLibrary
 {
-    public override List<Conditioned<Step>> Steps =>
+    public override List<AdoExpression<Step>> Steps =>
```
premun, line 201: @Meir017 here for instance.

Thank god for your doc generation as this might be easy to miss.

PR #459, opened 2025-06-14
Update libraries

premun:


premun:
Too many warnings about old frameworks not being supported

PR #528, opened 2026-04-10
Fix PipelineTrigger.Stages and Tags to use flat list instead of InclusionRule

kntng:
## Problem

`PipelineTrigger.Stages` and `PipelineTrigger.Tags` are typed as `InclusionRule?`, which serializes with include:/exclude: nesting. However, the Azure DevOps YAML schema only supports flat lists for both 
stages and tags in pipeline resource triggers — unlike branches which does support include/exclude.

### Generated (incorrect):

```yaml
 trigger:
   stages:
     include:
     - build
 ```

### Expected (per ADO schema):

```yaml
 trigger:
   stages:
   - build
  ```

## Fix

Changed Stages and Tags from `InclusionRule?` to `List<string>?` on `PipelineTrigger`.

References

 - Stage filters docs (https://learn.microsoft.com/en-us/azure/devops/pipelines/process/pipeline-triggers?view=azure-devops#stage-filters)
 - Tag filters docs (https://learn.microsoft.com/en-us/azure/devops/pipelines/process/pipeline-triggers?view=azure-devops#tag-filters)
 - YAML schema: resources.pipelines.pipeline (https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/resources-pipelines-pipeline?view=azure-pipelines) — shows stages: [ string ] and tags:
 [ string ]

### DIFF
--- a/src/Sharpliner/AzureDevOps/Model/PipelineTrigger.cs
+++ b/src/Sharpliner/AzureDevOps/Model/PipelineTrigger.cs
@@ -1,4 +1,5 @@
-﻿using System.Diagnostics.CodeAnalysis;
+﻿using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using Sharpliner.AzureDevOps.Expressions;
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Sharpliner/AzureDevOps/Model/PipelineTrigger.cs [hunk 1]
```diff
@@ -22,18 +23,20 @@ public record PipelineTrigger
...
 
     /// <summary>
-    /// List of tags to evaluate for trigger event
+    /// List of tags to evaluate for trigger event.
+    /// See <see href="https://learn.microsoft.com/en-us/azure/devops/pipelines/process/pipeline-triggers?view=azure-devops#tag-filters">tag filters</see>.
     /// Optional, 2020.1 and greater
     /// </summary>
     [DisallowNull]
-    public InclusionRule? Tags { get; init; }
+    public List<string>? Tags { get; init; }
```
premun, line 31: Can there be `${{ if }}` in the tags/stages? If yes, this should be `ConditionedList<string>?` instead

premun (commented):
Thank you for the change, left a quick question there

src/Sharpliner/AzureDevOps/Model/PipelineTrigger.cs
kntng: @premun I get the error `A template expression is not allowed in this context` when attempting to validate a YAML with a condition expression so I do not believe so.

premun: Thanks for checking

premun:
@kntng thanks for the change, it will be shortly published in `1.8.11`

PR #536, opened 2026-05-20
Fix #534: preserve conditional expressions on scalar AdoExpression properties

### LINKED ISSUE #534: JobBase.TimeoutInMinutes drops conditional Timeout expression
### Problem

`JobBase.Timeout` is an `AdoExpression<TimeSpan>?`, so callers can set it with a conditional expression. However, `JobBase.TimeoutInMinutes` converts the `Timeout` to minutes by reading only `Timeout.Definition`, which loses the condition when the YAML is serialized.

In practice, this makes a conditional timeout become unconditional YAML.

### Repro

```csharp
var timeoutCondition = If.In("parameters.rolloutInfra", "Test", "PPE");

var job = new Job("Ev2_Deployment", "Ev2 Deployment")
{
    Timeout = timeoutCondition.Value(TimeSpan.FromMinutes(60)),
};
```

### Expected YAML

The timeout should stay conditional, for example:

```yaml
- job: Ev2_Deployment
  displayName: Ev2 Deployment
  ${{ if in(parameters.rolloutInfra, 'Test', 'PPE') }}:
    timeoutInMinutes: 60
```

### Actual YAML

The condition is dropped and the timeout becomes unconditional:

```yaml
- job: Ev2_Deployment
  displayName: Ev2 Deployment
  timeoutInMinutes: 60
```

### Notes

`JobBase.Timeout` is marked `[YamlIgnore]`, so serialization relies on `JobBase.TimeoutInMinutes`:

```csharp
public AdoExpression<int>? TimeoutInMinutes => Timeout?.Definition != null ? (int)Timeout.Definition.TotalMinutes : null;
```

That conversion preserves the `TimeSpan` value but not the `AdoExpression` condition. `Step.Timeout` appears to handle conditional timeouts differently by materializing a conditional `TimeoutInMinutes`; `JobBase` likely needs similar behavior.

eerhardt:
Fixes #534.

## Why

`JobBase.TimeoutInMinutes` (and `CancelTimeoutInMinutes`, `Step.Timeout`, and in fact *every* scalar `AdoExpression<T>` property — `ContinueOnError`, `Enabled`, enums, etc.) silently dropped or threw on conditional `${{ if ... }}` expressions assigned via `If.X.Value(...)`. The issue reproduced against the published 1.8.11 package and made it impossible to write things like "use a longer timeout in production".

## Approach

Conditional `AdoExpression` already worked for mapping- and list-valued properties because the condition could be expressed *inside* the value (`pool:` then `${{ if }}:` then `name: foo`). Scalars don't have that luxury — the only YAML AzDO accepts is to lift the condition *above* the property:

```yaml
${{ if eq(...) }}:
  timeoutInMinutes: 60
```

But by the time a value-position converter runs, YamlDotNet has already emitted `timeoutInMinutes:`, so the fix has to live one level up — at the property-descriptor layer. The new `ConditionalScalarTypeInspector` (a YamlDotNet `ITypeInspector` decorator) detects conditional scalar `AdoExpression<T>` properties and replaces them with N synthetic sibling descriptors, one per branch, each named `${{ if ... }}` with a one-key inner mapping `{ realName: value }`. Full `If` / `ElseIf` / `Else` chains are walked from the synthetic root.

`JobBase.TimeoutInMinutes` / `CancelTimeoutInMinutes` and `Step.Timeout` need to convert `TimeSpan -> int minutes` while preserving the conditional tree. That's done with a new `AdoExpressionConversion.ConvertScalar<TSrc,TDst>(source, convert)` helper that mirrors the tree applying the converter to each leaf.

## Tests

Added `ConditionalScalarPropertyTests` with 11 cases covering `Job.Timeout`, `Job.CancelTimeout`, `Step.Timeout`, and `Step.ContinueOnError`, each exercised with a single condition, `If`/`Else`, and `If`/`ElseIf`/`Else` chains, plus an unconditional regression case. All pass; the public-API snapshot test also passes.

## Notes for reviewers

- `AdoExpression<T>` gained an `internal static CreateEmpty()` factory so `ConvertScalar` can construct synthetic root nodes without widening the public surface.
- The inspector only fires for properties whose declared type is `AdoExpression<TScalar>` where `TScalar` is a primitive, enum, string, `TimeSpan`, or `decimal`. Mapping- and list-valued conditional properties go through the existing `IYamlConvertible.Write` path unchanged.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/src/Sharpliner/AzureDevOps/Expressions/AdoExpression.cs
+++ b/src/Sharpliner/AzureDevOps/Expressions/AdoExpression.cs
@@ -101,6 +101,14 @@ public abstract record AdoExpression : IYamlConvertible
 
     internal abstract void WriteInternal(IEmitter emitter, ObjectSerializer nestedObjectSerializer);
 
+    /// <summary>
+    /// Returns the boxed value of <see cref="AdoExpression{T}.Definition"/> for this node, or
+    /// <see langword="null"/> if this node has no direct value (e.g. it is a parent that holds
+    /// child <see cref="Definitions"/> only). Used by serialization helpers that need to inspect
+    /// the value without knowing the generic parameter.
+    /// </summary>
+    internal virtual object? GetDefinitionValue() => null;
+
     internal virtual void SetIsList(bool isList)
     {
         IsList = isList;
@@ -327,6 +335,8 @@ public record AdoExpression<T> : AdoExpression
     /// </summary>
     internal T? Definition { get; }
 
+    internal override object? GetDefinitionValue() => Definition;
+
     internal AdoExpression(T definition, IfCondition condition) : base(condition)
     {
         Definition = definition;
@@ -355,6 +365,12 @@ public record AdoExpression<T> : AdoExpression
     protected AdoExpression() : base((IfCondition?)null)
     {
     }
+
+    /// <summary>
+    /// Internal factory for creating an empty <see cref="AdoExpression{T}"/> to act as a
+    /// synthetic root for an If/ElseIf/Else chain.
+    /// </summary>
+    internal static AdoExpression<T> CreateEmpty() => new();
 
     /// <summary>
     /// Starts a new <c>${{ if (...) }}</c> section.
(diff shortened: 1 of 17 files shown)


premun (approved):

premun:
Thanks @eerhardt. In about 15 minutes there will be `1.8.12` out

PR #542, opened 2026-05-21
Split Sharpliner into Sharpliner and Sharpliner.Core

eerhardt:
Sharpliner currently ships a single NuGet that drags `Microsoft.Build.Framework` and `Microsoft.Build.Utilities.Core` into every consumer, even when the consumer only wants the API to generate YAML programmatically (for example from a console app or a custom tool). This PR splits the assembly into two NuGets so users can opt out of the MSBuild dependency.

## Approach

- New `Sharpliner.Core` project containing the public Azure DevOps + GitHub Actions APIs, the serializer, and the configuration plumbing. No MSBuild references.
- Existing `Sharpliner` project keeps the MSBuild task (`PublishDefinitions`), `build/Sharpliner.props`, `build/Sharpliner.targets`, and now takes a `PackageReference` on `Sharpliner.Core`.
- `Sharpliner.nupkg` continues to bundle `Sharpliner.Core.dll`, `YamlDotNet.dll`, and `OneOf.dll` into its `lib/` folder via the existing `IncludeDependenciesInNupkg` target, so the MSBuild task loads them exactly as before. Consumers of the MSBuild integration see no behavior change.
- `RootNamespace` on the new project is `Sharpliner`, so all existing namespaces (`Sharpliner.AzureDevOps`, `Sharpliner.GitHubActions`, ...) are preserved.

## What actually changed in the moved files

155 files moved from `src/Sharpliner` to `src/Sharpliner.Core`. Of those, only a handful needed edits:

- Seven definition classes (`PipelineDefinition`, `PipelineDefinitionCollection`, `TemplateDefinition`, `TemplateDefinitionCollection`, `WorkflowDefinition`, plus two small references) now call a new `SharplinerDefaults.GetDefaultHeader` helper in Core instead of reaching into `SharplinerPublisher`, which still lives in the upper assembly.
- `AzureDevOpsDefinition.cs` uses the literal string `"FailIfChanged"` instead of `nameof(PublishDefinitions.FailIfChanged)` (Core cannot reference the MSBuild task type).
- `DotNetRestoreCoreCliTask.cs` swaps a `<see cref="...LoggerVerbosity"/>` doc reference for a `<c>` block since Core no longer references `Microsoft.Build.Framework`.
- `Properties/AssemblyInfo.cs` adds `[assembly: InternalsVisibleTo("Sharpliner")]`.

## Things worth a careful look

- `InternalsVisibleTo("Sharpliner")` on `Sharpliner.Core` is required because `SharplinerPublisher` still uses `SharplinerConfiguration.Current`, `SharplinerConfiguration.ConfigureInternal()`, and `DefaultSharplinerConfiguration`. I considered making them public or moving the publisher state into Core, but those are bigger redesigns and weren't necessary for the split. Keeping the IVT preserves the exact pre-split behavior and public API.
- The public API export test was split into two `*.verified.txt` files, one per assembly, so changes to either surface are tracked independently.
- `docs/AzureDevOps/GettingStarted.md` adds a short tip block explaining when to reach for `Sharpliner` vs `Sharpliner.Core`.
- Build scripts (`eng/scripts/build-dependencies.{ps1,sh}`) and `Sharpliner.slnx` were updated to pack and include the new project.

## Validation

- `dotnet build src/Sharpliner/Sharpliner.csproj` succeeds.
- `dotnet pack` produces `Sharpliner.43.43.43.nupkg` (with `Sharpliner.Core.dll`, `YamlDotNet.dll`, `OneOf.dll` bundled per-TFM in `lib/`) and `Sharpliner.Core.43.43.43.nupkg`.
- `dotnet build eng/Sharpliner.CI/Sharpliner.CI.csproj` (the meta project that consumes Sharpliner via NuGet) builds against the new packages.
- Both `PublicApisHaventChanged` tests pass after the verified files were updated for the split.


premun:
Thanks @eerhardt

